PR #242, opened 2024-12-15
unify opencompass

Yunnglin:
- unify opencompass and vlmeval output path
- output path lazy initialize
- replace `|` with `union` for python 3.10 ealier

### DIFF
--- a/evalscope/utils/io_utils.py
+++ b/evalscope/utils/io_utils.py
@@ -150,3 +150,13 @@ def json_to_dict(json_file) -> dict:
             raise e
 
     return stream
+
+
+def are_paths_same(path1, path2):
+    """
+    Check if two paths are the same.
+    """
+    real_path1 = os.path.realpath(os.path.abspath(path1))
+    real_path2 = os.path.realpath(os.path.abspath(path2))
+
+    return real_path1 == real_path2
--- a/evalscope/utils/io_utils.py
+++ b/evalscope/utils/io_utils.py
@@ -156,7 +156,7 @@ def are_paths_same(path1, path2):
     """
     Check if two paths are the same.
     """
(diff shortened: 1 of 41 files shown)


evalscope/utils/io_utils.py
```diff
@@ -0,0 +1,162 @@
...
+            raise e
+
+    return stream
+
+
+def are_paths_same(path1, path2):
+    """
+    Check if two paths are the same.
+    """
+    real_path1 = os.path.realpath(os.path.abspath(path1))
```
wangxingjun778: Add `os.path.expanduser()` to expand the path like `~/.cache/path/to/xxx`

examples/example_eval_swift_openai_api.py
```diff
@@ -47,16 +47,16 @@ def run_swift_eval():
     task_cfg = dict(
         eval_backend='OpenCompass',
         eval_config={
-            'datasets': ['winogrande'],
+            'datasets': ['gsm8k', 'ARC_c'],
             'models': [
                 {
-                    'path': 'qwen2-7b-instruct',  # Please make sure the model is deployed
+                    'path': 'qwen2.5',  # Please make sure the model is deployed
```
wangxingjun778: The `path` should be aligned with MS/HF model_name, e.g. for Qwen/Qwen2.5-7B-Instruct, the path should be `Qwen2.5-7B-Instruct`

evalscope/utils/io_utils.py
Yunnglin: done

examples/example_eval_swift_openai_api.py
Yunnglin: done

PR #260, opened 2024-12-25
Fix: sglang API response does not contain 'object' field.

tghfly:
Fix: sglang API response does not contain 'object' field.


Yunnglin:
做了一点修改，默认先解析usage

wangxingjun778 (approved):

PR #370, opened 2025-03-06
Compat reasoning model and support filter

Yunnglin:


evalscope/models/server_adapter.py
```diff
@@ -123,28 +144,32 @@ def send_request(self, request_json: dict) -> dict:
     def _collect_stream_response(self, response_stream: List[ChatCompletionChunk]) -> ChatCompletion:
```
wangxingjun778, line 144: collecting `usage` is recommended when stream_options is {'include_usage': True}, an example could be:

```
if chunk.usage:
    usage = chunk.usage.to_dict()
```

PR #634, opened 2025-06-06
Fix in benchmark, when the number of dataset index is less than parallel , the parallel will be insufficient.

xcode03:
e.g. If the dataset contains 700 records and the benchmark parallel argument is set to 1000, the actual parallel of the benchmark is only 700. 
And the number of records set by the number argument is not reached after the first round of parallel requests, the benchmark will end early.


Yunnglin (approved):

Yunnglin:
LGTM. Thanks for your PR!

PR #663, opened 2025-06-20
make sure the stream parameter is included in the request_json

Su-yj:
Make sure the `stream` parameter is included in the `request_json`, because some privately deployed large model services set the `stream` parameter to `True` by default.


Yunnglin (approved):

Yunnglin:
LGTM!

PR #745, opened 2025-07-29
Fix: Incorrect keyword argument in call to csv_to_list()

Zhuzhenghao:
### Summary

This PR fixes a `TypeError` caused by an incorrect keyword argument (`file_path=...`) passed to the `csv_to_list()` function in `summarizer.py`.

### Problem

The current call:

```python
csv_to_list(file_path=summary_file_path)
````

raises the following error at runtime:

```
TypeError: csv_to_list() got an unexpected keyword argument 'file_path'
```

This happens because the `csv_to_list()` function only accepts a single positional argument:

```python
def csv_to_list(csv_file)
```

### Fix

Changed the call to:

```python
csv_to_list(summary_file_path)
```

which correctly matches the function's signature.

### Impact

This fix allows the summarizer to work as expected when generating reports via `Summarizer.get_report_from_cfg(...)`.

No other functionality is affected.


Yunnglin:
LGTM! Thanks for your PR!

PR #815, opened 2025-09-04
add cmmlu and gsm8k

akaihaoshuai:
add cmmlu and gsm8k


Yunnglin:
Thank you for your PR! There are a few minor issues that need to be fixed:

1. Run the following commands to fix lint issues:
   ```bash
   pip install pre-commit
   pre-commit install
   pre-commit run --all-files
   ```

2. Change `yield [prompt]` to `yield prompt`

3. I couldn't find the `haonan-li/cmmlu` dataset on ModelScope. Is there an issue with the data source?

akaihaoshuai:
fixed

Yunnglin:
please change `yield [message]` to
```python
if self.query_parameters.apply_chat_template:
    message = self.create_message(prompt)
    yield [message]
else:
    yield prompt
```
in gsm8k and openqa

PR #873, opened 2025-10-14
Datasets: prompt tokens count bug fixed

Aktsvigun:
Hey team!

I've noticed a bug in the custom datasets implementation. Namely, when taking the inputs from the dataset (regardless of the format - `openqa` / `custom`/ etc.), rather than calculating the number of **tokens** in the input and checking whether it meets the min/max requirements, it calculates the number of symbols. 

I understand this may be equivalent for Chinese (since likely one symbol is 1 token), but incorrect for Latin/Slavic characters. I've implemented a fix so that the number of tokens is calculated with the tokenizer.

Looking forward to your approval!

### DIFF
--- a/evalscope/perf/plugin/datasets/base.py
+++ b/evalscope/perf/plugin/datasets/base.py
@@ -104,5 +104,5 @@ class DatasetPluginBase:
             prompt_length = len(prompt)
         else:
             prompt_length = len(self.tokenizer.encode(prompt))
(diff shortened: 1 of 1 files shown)


Yunnglin:
Thank you for your pull request! Since the tokenizer is an optional parameter, I have adjusted the logic to accommodate both scenarios: whether the parameter is provided or not.

Aktsvigun:
Thank you!

PR #964, opened 2025-11-07
[ADD] Implement metric aggregation pass@k and vote@k #387

xin8coder:
**New Aggregation Classes**
1. MeanPassAtK: Computes pass@k by taking the maximum score per group 
3. MeanVoteAtK: Computes vote@k using the most frequent prediction

**Key Features**
1. Dynamic k-value calculationfrom group distribution
2. Multi-metric supportwith automatic field propagation 
4. Integrated mean aggregationfor all derived metrics

**Usage Note**
The aggregation functions preserve all original benchmark metrics. For example, when the original benchmark metric is mean_acc, using the mean_and_vote_at_kaggregator with repeats=8will automatically compute and add the mean_acc_vote@8metric while preserving the original mean_accmetric.

**Ensures:**
Backward compatibility- All original metrics remain unchanged
Metric enrichment- Automatically adds group-based aggregated metrics
Flexibility- Users can choose between original or aggregated metrics for analysis

**Usage:**

To use the new aggregation methods, specify the aggregator name in the dataset_args configuration:

`from evalscope import TaskConfig, run_task

task_cfg = TaskConfig(
    model='Qwen3-4B-Instruct',
    api_url='http://127.0.0.1:8000/v1/chat/completions',
    eval_type='server',
    datasets=['gpqa_diamond', 'aime25', 'ifeval'],
    dataset_args={
        'gpqa_diamond': {'aggregation': 'mean_and_vote_at_k'},
        'aime25': {'aggregation': 'mean_and_pass_at_k'},
        'ifeval': {'aggregation': 'mean_and_pass_at_k'}
    },
    eval_batch_size=256,
    ignore_errors=True,
    repeats=8,  # Determines k-value
    generation_config={
        'max_tokens': 10000,
        'temperature': 0.7,
        'top_p': 0.8,
        'top_k': 20
    },
    timeout=60000,
    stream=True,
    limit=10
)

run_task(task_cfg=task_cfg)`

result

### DIFF
--- a/evalscope/metrics/metric.py
+++ b/evalscope/metrics/metric.py
@@ -1,4 +1,5 @@
 import json
+import numpy as np
 import os
 from collections import defaultdict
 from typing import Dict, List
@@ -6,7 +7,7 @@ from typing import Dict, List
 from evalscope.api.metric import Aggregator, AggScore, Metric, SampleScore, SingletonMetric, T2IMetric
 from evalscope.api.registry import register_aggregation, register_metric
 from evalscope.utils.import_utils import check_import
-from .metrics import mean
+from .metrics import calculate_pass_at_k, mean
 
 
 def normalize_text(text: str) -> str:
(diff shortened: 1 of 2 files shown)


xin8coder (commented):
fixed.

PR #999, opened 2025-11-24
fix: fix --no-test-connection args action

ljwh:
Change-Id: If7b886867e5d77b37fe7add3388b87780174eeae

Hi, the args --no-test-connection for perf test does not work currently, because the default value is False and action is store_false, the final value is False no matter how you set the config.

Replace store_false to store_true to solve the problem.


ljwh:
@Yunnglin PTAL

PR #1045, opened 2025-12-04
[feature] add lazy model

Secbone:
When using the cache for review or reporting, the original model is not always required to be loaded. Adding LazyModel ensures that the model is only loaded when necessary, speeding up the evaluation process and freeing up unnecessary resources.


Yunnglin:

```shell

pip install pre-commit

pre-commit install

pre-commit run --all-files

```

PR #1083, opened 2025-12-15
fix: Output token throughput and Total token throughput on Concurrency 1

cdpath:
I encountered the same issue as #1076.

Here is a simple fix.


Yunnglin (approved):

PR #1214, opened 2026-03-09
feat(perf): add --same-request option for benchmark

yonlunwu:
When enabled, all concurrent requests will send the same prompt (only for dataset mode).

This is useful for simulating cache hit scenarios with different input lengths. By reusing a single prompt, you can easily measure the performance of cache hits without preparing multiple prompts of different lengths. It also helps with model warm-up testing.


Yunnglin:
This feature's necessity can be further discussed. Currently, you can use `random` dataset and configure the `prefix-length` to ensure that the request is the same, thereby testing cache hits.

Yunnglin:
Hi @yonlunwu, thanks for the contribution!

After discussion, we've decided not to merge this feature for now. The same effect (sending identical prompts to simulate cache hits) can already be achieved by using `--dataset random` with a fixed `--prefix-length` configuration. Adding a separate `--same-request` flag would introduce a redundant path for the same use case.

Additionally, there are a couple of implementation issues noted (the `not args.dataset` condition would never trigger due to the default value, and there's a merge conflict with `main`).

Closing this PR — feel free to reopen or discuss if you have a different use case in mind that isn't covered by the existing approach. Thanks again!

PR #1221, opened 2026-03-10
Replace eval() with ast.literal_eval() in ParseStrArgsAction

RinZ27:
## What

Switch from `eval()` to `ast.literal_eval()` in `ParseStrArgsAction` within `evalscope/arguments.py`. 

## Why

The current implementation uses `eval()` to parse CLI arguments, which introduces a potential code injection vulnerability. Since `model-args` are often complex structures (dicts, lists), switching to `ast.literal_eval()` provides a secure way to evaluate these Python literals without the risk of arbitrary code execution. 

## Type

- [x] Bug fix / Security hardening
- [ ] Refactor

## Testing

Verified that `ast.literal_eval()` correctly parses integers, booleans, lists, and dictionaries while preventing any function calls or malicious payloads.

### DIFF
--- a/evalscope/arguments.py
+++ b/evalscope/arguments.py
@@ -1,5 +1,6 @@
 # flake8: noqa: E501
 import argparse
+import ast
 import json
 
 from evalscope.constants import EvalBackend, JudgeStrategy, ModelTask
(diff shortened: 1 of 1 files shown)


Yunnglin (approved):

PR #1242, opened 2026-03-20
print request id

strenuous-life:
In order to facilitate the identification of backend issues, such as sglang, the request_id should be printed.


Yunnglin:
Thank you for your PR, I have modified the implementation to output the requested ID to the predictions file instead of printing too much log.

PR #1243, opened 2026-03-20
[Fix] Use seeded random.Random for dataset shuffling

pcabriada:
Python 3.12 removed the deprecated second argument from random.shuffle(). LocalDataLoader.load() and DictDataLoader.load() both passed self.seed as the second argument, causing TypeError on Python 3.12+.

Replace with random.Random(seed).shuffle(dataset) when a seed is provided, falling back to unseeded random.shuffle() when seed is None.

RemoteDataLoader is unaffected (uses HuggingFace Dataset.shuffle(seed=...)).

Added tests/test_loader_shuffle.py covering seeded shuffle, unseeded shuffle, reproducibility, and a guard confirming the original call fails on 3.12+.


Yunnglin (approved):

PR #1266, opened 2026-04-01
feat: save reasoning content into review result

jiaohuix:
Reason
Evaluating reasoning-intensive models is computationally expensive and time-consuming.However, the original pipeline does not persist the model's internal reasoninging process during evaluation. As a result, valuable reasoning content is discarded after each inference, leading to wasted computation when re-analyzing or reviewing results.
Changes
Add reasoning field to the ReviewResult Pydantic model
Extract and store model reasoning content from both reasoning and reasoning_content response fields
Safely handle empty responses and metadata to avoid index errors
Persist reasoning content in the review cache to avoid redundant model inference
This improvement preserves useful reasoning logs without re-computing evaluations, improving efficiency in result analysis and debugging.

### DIFF
--- a/evalscope/api/evaluator/cache.py
+++ b/evalscope/api/evaluator/cache.py
@@ -196,11 +196,8 @@ class CacheManager:
             The saved review result object
         """
         cache_file = self.get_review_cache_path(subset)
-        output = task_state.output
         # Convert score and state to serializable review result
         review_result = ReviewResult.from_score_state(sample_score, task_state, save_metadata)
-        # Save reasoning content into review result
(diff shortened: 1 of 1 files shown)


Yunnglin:
Currently, the reasoning content is saved and output to the `predictions` directory as a jsonl file, with the field name being `reasoning`, while the review process does not refer to the reasoning. So I think this PR isn’t necessary and can be closed.

image

jiaohuix:
Thanks for your review! I didn't notice this before. You're right, I'll close this PR.

PR #1282, opened 2026-04-10
fix: support unicode operators (≤, ≥) in SLA criterion

yonlunwu:


yonlunwu:
考虑到两个都是小 patch 且都是 sla 相关，关闭该 pr，统一到 https://github.com/modelscope/evalscope/pull/1283

PR #1295, opened 2026-04-15
[Benchmark] Add TIR-Bench

Yunnglin:


ninghongbo123:
@Yunnglin 
Hello,
I noticed that the tir-bench setup in the paper uses a code interpreter or tool call, which is different from the current evalcope implementation.
How are models like Qwen 3.5 evaluated on this task now?
Thanks !

PR #1348, opened 2026-05-18
core: fix build_docker_images check to always subset dataset first

liguodongiot:
when build_docker_images is false, update image metadata exception.


Yunnglin (approved):

PR #1363, opened 2026-05-22
[Fix] Fix ineffective repeats in DatasetDict.from_dataset

we1sper:
Fix `DatasetDict.from_dataset` to actually repeat samples instead of just taking more distinct samples.

https://github.com/modelscope/evalscope/blob/05649760fafd71d27fabc0c5e58559aedfd89202/evalscope/api/dataset/dataset.py#L336-L348

Previously, `repeats` was applied as `total_limit = limit * repeats`, which simply sliced more distinct samples from the dataset. This violates the semantics of `repeats`.


Yunnglin (approved):

PR #1370, opened 2026-05-25
fix(perf): round int fields after sla multi-run averaging

qiumuyang:
## Problem

When `--sla-num-runs >= 2`, `db_util.average_results` computes arithmetic means over all numeric fields. 

Int-typed count fields (`concurrency`, `total_requests`, `succeed_requests`, `failed_requests`) can produce fractional values (e.g. `(390+391+391)/3 = 390.67`) that Pydantic strict-int validation rejects with a `ValidationError`, crashing the SLA tuner.

## Solution

Add `@field_validator(mode='before')` on the four int fields of `BenchmarkSummary` that rounds float inputs back to int. The rounding happens at the schema layer where the int contract is declared, keeping `db_util.average_results` schema-agnostic.

### DIFF
--- a/evalscope/perf/utils/perf_models.py
+++ b/evalscope/perf/utils/perf_models.py
@@ -16,7 +16,7 @@ Public classes
 from __future__ import annotations
 
 import json
-from pydantic import BaseModel, ConfigDict, Field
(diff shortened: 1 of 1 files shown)


Yunnglin (approved):

PR #1381, opened 2026-05-28
fix(scicode): read assistant text blocks

he-yufeng:
﻿## Summary
- use ChatMessage.text when collecting SciCode assistant solutions
- keep the existing markdown code-fence extraction behavior unchanged
- add regression coverage for assistant messages whose content is stored as text blocks

Fixes #1380.

## To verify
- python -m py_compile evalscope\benchmarks\scicode\util.py tests\benchmark\test_scicode_util.py
- python -m pytest tests\benchmark\test_scicode_util.py -q
- python -m flake8 evalscope\benchmarks\scicode\util.py tests\benchmark\test_scicode_util.py
- python -m yapf --diff evalscope\benchmarks\scicode\util.py tests\benchmark\test_scicode_util.py
- git diff --check

PR #1400, opened 2026-06-05
fix(scicode): read assistant text blocks

he-yufeng:
## Summary

SciCode generated-code extraction now reads assistant text blocks through `message.text` instead of casting `message.content` to `str`.

`ChatMessageAssistant.content` can be a list of typed content blocks. Casting that list to `str` leaves Python object/list syntax in the extracted code, which breaks downstream SciCode execution. The existing message abstraction already exposes the flattened text, so the fix uses that path directly.

## Validation

- `python -m pytest tests\benchmark\test_scicode_util.py -q`
- `python -m py_compile evalscope\benchmarks\scicode\util.py tests\benchmark\test_scicode_util.py`
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/evalscope/benchmarks/scicode/util.py
+++ b/evalscope/benchmarks/scicode/util.py
@@ -41,11 +41,7 @@ def get_generated_code(state: TaskState) -> list[str]:
         If the number of assistant messages does not match the number of subproblems in the task metadata.
 
     """
-    assistant_messages = [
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this because upstream/main already contains the same SciCode text-block fix and regression test via #1381. Keeping this open would just duplicate the merged change.

PR #1440, opened 2026-06-26
(feat) Parallelize CPU-bound perf request generation

### LINKED ISSUE #1433: 性能测试时候，生成请求阶段用多核cpu并行生成
## 功能描述 / Feature Description

性能测试生成请求阶段用多核cpu并行

## 需求背景 / Background

高并发长输入要2 3分钟才能生成完请求，我的测试机器有200多个cpu核心，不用多核cpu浪费了

## 预期行为 / Expected Behavior

自动检测cpu核心数量，用cpu核心数量并行生成请求

## 其他信息 / Additional Information

无

haoruilee:
## Summary

Add shared CPU-bound dataset/request generation workers for perf benchmarks.

This now uses the existing `num_workers` concept instead of adding a separate request-generation-only option. The top-level perf CLI option is `--num-workers`; legacy `multi_turn_args.num_workers` is still accepted only as a compatibility bridge when top-level `num_workers` is left at `0`.

Closes #1433

## Parallel Acceleration Algorithm

1. Worker count is configured by top-level `--num-workers`:

- `0`: auto mode.
- `1`: force serial generation.
- `>1`: explicitly use up to that many worker processes, capped by the item count.

2. Shared worker resolution lives in `evalscope/perf/utils/worker_util.py` and is used by both `benchmark.py` and `swe_smith` live construction:

```python
amortized_workers = max(1, total_count // 128)
workers = min(available_cpu_count, total_count, 32, amortized_workers)
```

`available_cpu_count` honors Linux CPU affinity via `os.sched_getaffinity(0)` before falling back to `os.cpu_count()`.

3. Random dataset auto mode has an additional dataset-level gate because spawn is not free:

- Explicit `--num-workers > 1` is honored.
- `--tokenize-prompt` does not auto-enable multiprocessing because that path is already cheap.
- Auto multiprocessing requires `prompt_length_estimate >= 8192`.
- Auto multiprocessing also requires `prompt_length_estimate * total_count >= 4 * 1024 * 1024` token-work units.

This intentionally avoids regressions on small/medium workloads where process startup, tokenizer reload, and IPC dominate.

4. Process start mode is `spawn`, not `fork`.

Each random-generation worker initializes process-local state through `ProcessPoolExecutor(initializer=...)`:

- `TOKENIZERS_PARALLELISM=false` is set in the child.
- A fresh tokenizer is loaded in each child when text prompt decoding is needed.
- Tokenizer objects are not reused across a fork boundary.
- Worker state is held in a `_RandomGenerationWorkerState` dataclass instead of a mutable global dict / copy-on-write relay.

5. Random generation itself remains deterministic and order-stable:

- The parent pre-samples the full per-item plan: `(input_len, offset, seed)`.
- Workers receive independent tasks: `(index, input_len, offset, seed)`.
- The parent fills the result array by `index`, so warmup remains first and request order remains stable.
- Missing indexes raise immediately.
- Token mismatch accounting is aggregated after worker results return.

6. `swe_smith` live construction now uses the same top-level worker resolver. It resolves workers after candidate work items are known, so the resolver is based on CPU work item count rather than requested output count.

## Benchmark Evidence

Command:

```bash
python3 examples/perf/benchmark_random_request_generation.py \
  --numbers 128 512 \
  --prompt-lengths 2048 8192 \
  --workers 1 2 4 8 0 \
  --tokenizer-path Qwen/Qwen2.5-0.5B-Instruct
```

Environment: local 32-core CPU affinity, Python 3.12, no HTTP requests, raw text prompts (`--apply-chat-template` disabled).

| number | prompt_length | configured_workers | resolved_workers | seconds | req/s |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 128 | 2048 | 1 | 1 | 5.519 | 23.19 |
| 128 | 2048 | 2 | 2 | 7.685 | 16.66 |
| 128 | 2048 | 4 | 4 | 10.466 | 12.23 |
| 128 | 2048 | 8 | 8 | 16.939 | 7.56 |
| 128 | 2048 | 0 | 1 | 3.437 | 37.24 |
| 128 | 8192 | 1 | 1 | 6.543 | 19.56 |
| 128 | 8192 | 2 | 2 | 9.224 | 13.88 |
| 128 | 8192 | 4 | 4 | 10.488 | 12.20 |
| 128 | 8192 | 8 | 8 | 16.950 | 7.55 |
| 128 | 8192 | 0 | 1 | 5.998 | 21.34 |
| 512 | 2048 | 1 | 1 | 10.426 | 49.11 |
| 512 | 2048 | 2 | 2 | 11.683 | 43.82 |
| 512 | 2048 | 4 | 4 | 11.752 | 43.57 |
| 512 | 2048 | 8 | 8 | 17.470 | 29.31 |
| 512 | 2048 | 0 | 1 | 10.026 | 51.07 |
| 512 | 8192 | 1 | 1 | 20.339 | 25.17 |
| 512 | 8192 | 2 | 2 | 17.174 | 29.81 |
| 512 | 8192 | 4 | 4 | 14.669 | 34.90 |
| 512 | 8192 | 8 | 8 | 17.479 | 29.29 |
| 512 | 8192 | 0 | 4 | 14.331 | 35.73 |

Interpretation: spawn overhead is material. Auto mode now stays serial where multiprocessing is slower, and enables 4 workers for the 512x8192 case where it improves wall-clock time from 20.339s to 14.331s (~1.42x).

## Validation

- `pytest tests/perf/test_request_generation.py -q`
- `python3 examples/perf/benchmark_random_request_generation.py --numbers 128 512 --prompt-lengths 2048 8192 --workers 1 2 4 8 0 --tokenizer-path Qwen/Qwen2.5-0.5B-Instruct`
- `make lint`
- `pytest tests/cli/test_all.py::TestRun::test_ci_lite -v -s -p no:warnings`

### DIFF
--- a/evalscope/perf/arguments.py
+++ b/evalscope/perf/arguments.py
@@ -187,6 +187,14 @@ class Arguments(BaseArgument):
     in_flight_task_multiplier: int = 2
     """Max scheduled tasks = parallel * this multiplier."""
 
+    request_generation_workers: int = 0
+    """Number of worker processes for CPU-bound request generation.
+
+    Set to 0 for auto-detection based on the current CPU affinity, 1 to force
+    serial generation, or >1 to explicitly choose the worker count. Only
+    dataset plugins that advertise parallel generation use this setting.
+    """
+
     # Logging and debugging
     log_every_n_query: int = 100
     """Log every N queries."""
@@ -430,6 +438,13 @@ class Arguments(BaseArgument):
     def _validate_in_flight_task_multiplier(cls, v):
         return max(v, 1) if v <= 0 else v
 
+    @field_validator('request_generation_workers', mode='after')
+    @classmethod
+    def _validate_request_generation_workers(cls, v: int) -> int:
+        if v < 0:
+            raise ValueError('--request-generation-workers must be >= 0')
+        return v
+
     # --- Model validator (cross-field logic) ---
 
     @model_validator(mode='after')
@@ -611,6 +626,15 @@ def add_argument(parser: argparse.ArgumentParser):
     parser.add_argument('--db-commit-interval', type=int, default=1000, help='Rows buffered before SQLite commit')
     parser.add_argument('--queue-size-multiplier', type=int, default=5, help='Queue maxsize = parallel * multiplier')
     parser.add_argument('--in-flight-task-multiplier', type=int, default=2, help='Max scheduled tasks = parallel * multiplier')  # noqa: E501
+    parser.add_argument(
+        '--request-generation-workers',
+        type=int,
+        default=0,
+        help=(
+            'Worker processes for CPU-bound request generation. '
+            '0=auto from CPU affinity, 1=serial, >1=explicit worker count.'
+        ),
+    )
 
     # Logging and debugging
     parser.add_argument('--log-every-n-query', type=int, default=100, help='Logging every n query')
--- a/evalscope/perf/arguments.py
+++ b/evalscope/perf/arguments.py
@@ -187,8 +187,8 @@ class Arguments(BaseArgument):
     in_flight_task_multiplier: int = 2
     """Max scheduled tasks = parallel * this multiplier."""
 
-    request_generation_workers: int = 0
-    """Number of worker processes for CPU-bound request generation.
+    num_workers: int = 0
+    """Number of worker processes for CPU-bound dataset/request generation.
 
     Set to 0 for auto-detection based on the current CPU affinity, 1 to force
     serial generation, or >1 to explicitly choose the worker count. Only
@@ -438,11 +438,11 @@ class Arguments(BaseArgument):
     def _validate_in_flight_task_multiplier(cls, v):
         return max(v, 1) if v <= 0 else v
 
-    @field_validator('request_generation_workers', mode='after')
+    @field_validator('num_workers', mode='after')
     @classmethod
-    def _validate_request_generation_workers(cls, v: int) -> int:
+    def _validate_num_workers(cls, v: int) -> int:
         if v < 0:
-            raise ValueError('--request-generation-workers must be >= 0')
+            raise ValueError('--num-workers must be >= 0')
         return v
 
     # --- Model validator (cross-field logic) ---
@@ -466,6 +466,13 @@ class Arguments(BaseArgument):
 
         # Validate sweep params (number/parallel/rate consistency)
         self._validate_sweep_params()
+
+        if (
+            self.num_workers == 0 and self.multi_turn_args is not None
+            and 'num_workers' in self.multi_turn_args.model_fields_set
+        ):
+            logger.warning('`multi_turn_args.num_workers` is deprecated. Please use top-level `--num-workers` instead.')
+            self.num_workers = self.multi_turn_args.num_workers
 
         return self
 
@@ -627,11 +634,11 @@ def add_argument(parser: argparse.ArgumentParser):
     parser.add_argument('--queue-size-multiplier', type=int, default=5, help='Queue maxsize = parallel * multiplier')
     parser.add_argument('--in-flight-task-multiplier', type=int, default=2, help='Max scheduled tasks = parallel * multiplier')  # noqa: E501
     parser.add_argument(
-        '--request-generation-workers',
+        '--num-workers',
         type=int,
         default=0,
         help=(
-            'Worker processes for CPU-bound request generation. '
+            'Worker processes for CPU-bound dataset/request generation. '
             '0=auto from CPU affinity, 1=serial, >1=explicit worker count.'
         ),
     )
@@ -753,7 +760,8 @@ def add_argument(parser: argparse.ArgumentParser):
             'Example: \'{"first_turn_length": 65000, "subsequent_turn_length": 500, '
             '"chars_per_token": 3.0}\'. '
             'Note: min_turns and max_turns are top-level --min-turns / --max-turns arguments '
-            '(per-conversation turn count is sampled from [min_turns, max_turns]).'
+            '(per-conversation turn count is sampled from [min_turns, max_turns]); '
+            'use top-level --num-workers for live construction parallelism.'
         ),
     )
     # yapf: enable
--- a/evalscope/perf/arguments.py
+++ b/evalscope/perf/arguments.py
@@ -468,7 +468,7 @@ class Arguments(BaseArgument):
         self._validate_sweep_params()
 
         if (
-            self.num_workers == 0 and self.multi_turn_args is not None
+            'num_workers' not in self.model_fields_set and self.multi_turn_args is not None
             and 'num_workers' in self.multi_turn_args.model_fields_set
         ):
             logger.warning('`multi_turn_args.num_workers` is deprecated. Please use top-level `--num-workers` instead.')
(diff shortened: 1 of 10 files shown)


Yunnglin (changes requested):
Thanks for tackling this — parallelizing the CPU-bound request generation phase is a meaningful optimization, and the overall design (pre-sampled plan + independent per-index workers) is well thought out. A few issues need to be resolved before this can be merged:

### 1. Unify worker parameter with `swe_smith` instead of introducing a new one

This PR adds `--request-generation-workers`, but `swe_smith` already uses `multi_turn_args.num_workers` for the exact same purpose (CPU-bound data construction parallelism). Having two separate parameters with different names and resolution logic for the same concept creates unnecessary divergence.

**Suggestion**: Promote `num_workers` to a top-level `Arguments` field and unify the worker resolution logic in `benchmark.py`, so that `random`, `swe_smith`, and future dataset plugins share the same mechanism.

### 2. Use `spawn` mode instead of `fork`

The PR hardcodes `fork` only, but `swe_smith` explicitly uses `spawn` with a clear rationale:

> Use spawn context to avoid fork-based deadlocks on Linux: the perf runner's parent process contains async loops, logger threads, tqdm monitors, and a possibly pre-warmed HF tokenizer (Rust Rayon threads). fork would inherit these threads' lock states and deadlock on Pool cleanup.

The `fork` risks here are identical:
- The perf runner already has asyncio event loops, logger threads, and tqdm monitors running
- HuggingFace tokenizer's internal Rust Rayon thread pool can deadlock/segfault after fork
- The `threading.active_count() != 1` gate in `_get_random_generation_context()` will almost certainly be True during actual perf runs (due to existing threads), causing the parallel path to silently degrade to serial

The worker function is already a top-level function with data passed via `initargs` — this is spawn-compatible by design. Switching should be straightforward.

### 3. Missing performance benchmark evidence

The test file only validates correctness (worker count resolution, fallback behavior, seed consistency). There is no wall-clock performance comparison proving that multiprocessing actually outperforms serial execution, given:
- Process spawn/fork overhead
- IPC serialization cost for each message (list/dict payloads)
- The `tokenize_prompt=True` path is already extremely fast (pure numpy slicing)

Please provide a benchmark script comparing serial vs parallel across different `(prompt_length, total_count, workers)` combinations, and include the results in the PR description.

### 4. Fragile global mutable state (`_RANDOM_GENERATION_STATE`)

Using a module-level `Dict` as a cross-process state relay is brittle:
- Under fork, it relies on CoW sharing, but GC-triggered page writes can degrade to full copies
- Under spawn, the dict is empty (module re-imported), making the "read existing → clear → update" logic in `_init_random_generation_worker` confusing
- The overall pattern is hard to reason about

**Suggestion**: With spawn mode, simply pass all required data through `initargs` directly — no global state needed. Alternatively, encapsulate in a `WorkerState` dataclass.

### 5. Thread-count gate won't fire in practice

`_get_random_generation_context()` checks `threading.active_count() != 1`, but within `benchmark.py`'s async context (asyncio internal threads, tqdm, logger are already running), this condition is almost always True → returns `None` → parallel path never triggers.

### 6. Minor code quality issues

- `benchmark.py`: `import os` should be at the top of the stdlib block, not inserted between `numpy` and `typing`
- `build_messages_parallel` L186: `process_context is not None` is a redundant check (already guaranteed non-None at L165)
- `_get_random_generation_context()` return type should be `Optional[multiprocessing.context.BaseContext]`, not `Optional[Any]`
- `self.number = total_count` in `build_messages_parallel` mutates instance state — potential side effect if `self.number` is referenced elsewhere

### Action Items

1. Unify the worker parameter with `swe_smith`'s `num_workers`
2. Switch to `spawn` mode (consistent with `swe_smith`)
3. Provide benchmark data demonstrating actual perf gain
4. Remove global mutable dict; pass state through `initargs`
5. Fix the thread-count gate logic

haoruilee:
Hi @Yunnglin ,

Thanks for the detailed review. I addressed the requested changes in the latest push:

1. Unified worker config around top-level `num_workers`.  `random` and `swe_smith` now share `resolve_dataset_generation_workers()`. `multi_turn_args.num_workers` is still accepted only as a deprecated fallback when top-level `num_workers` is omitted.

2. Switched random request generation to `spawn`.

3. Added benchmark tooling.  I added `examples/perf/benchmark_random_request_generation.py` to compare serial vs parallel request generation across prompt lengths, request counts, and worker counts. 

I've also tested it locally. Command used:

```
python3 examples/perf/benchmark_random_request_generation.py \
  --tokenizer-path Qwen/Qwen2.5-0.5B-Instruct \
  --numbers 128 512 \
  --prompt-lengths 2048 8192 \
  --workers 1 2 4 8 0 \
  --repeat 1
```

Environment: Python 3.12.3, 32 CPUs. `workers=0` means auto.

| total_count | prompt_len | workers=1 | workers=2 | workers=4 | workers=8 | auto |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 128 | 2048 | 4.736s | 7.715s | 10.403s | 17.094s | 3.512s, resolved 1 |
| 128 | 8192 | 6.124s | 9.087s | 10.428s | 16.923s | 5.951s, resolved 1 |
| 512 | 2048 | 10.592s | 11.527s | 11.458s | 17.243s | 10.070s, resolved 1 |
| 512 | 8192 | 20.154s | 17.280s | 14.292s | 17.965s | 14.153s, resolved 4 |

A larger long-prompt case:

| total_count | prompt_len | workers=1 | workers=2 | workers=4 | workers=8 | auto |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 1024 | 8192 | 40.047s | 28.150s | 20.271s | 19.850s | 20.031s, resolved 8 |

Explicit multiprocessing is slower on small/medium cases due to spawn/tokenizer startup overhead, so auto correctly stays serial there. It helps on long large workloads: `512 x 8192` improves from 20.154s to 14.153s, about
1.42x, and `1024 x 8192` improves from 40.047s to 20.031s, about 2.0x.

4. Reworked worker state.  The mutable global dict was replaced with a process-local `WorkerState` dataclass initialized via `ProcessPoolExecutor` initargs. This keeps the tokenizer/allowed-token state initialized once per spawned worker while passing only per-item tasks through `map`.

5. Fixed some minor code quality issues.

Validation:
- `pytest tests/perf/test_request_generation.py -q`
- `pytest tests/cli/test_all.py::TestRun::test_ci_lite -v -s -p no:warnings`
- `make lint`

Yunnglin (approved):
LGTM — all prior review items addressed, code deduplication applied.

### Summary

- Unified worker config via top-level `--num-workers` (shared with `swe_smith`)
- Switched to `spawn` mode (consistent with `swe_smith`)
- Benchmark evidence provided (1.3x speedup on 1024×8192)
- Global mutable dict replaced with `_RandomGenerationWorkerState` dataclass + `ProcessPoolExecutor(initializer=...)`
- Thread-count gate removed; replaced with prompt-length × token-work threshold
- Serial/worker logic deduplicated via `_build_one_random_message` shared pure function

### Minor follow-up (non-blocking)

- Consider extracting L20-146 worker infrastructure into `_random_generation_worker.py` in a follow-up to improve file navigability (431 lines, class definition starts at L149)

PR #1461, opened 2026-07-07
Add PerspectiveGap benchmark adapters

WhymustIhaveaname:
# Summary

Adds EvalScope adapters for [PerspectiveGap](https://arxiv.org/abs/2606.08878), a benchmark for multi-agent orchestration prompting and role-specific context routing.

- Registers `perspective_gap_role_assignment` and `perspective_gap_prompt_writing`.
- Loads the ModelScope mirror `evalscope/PerspectiveGap` by default.
- Scores outputs through the official `perspective_gap.scoring` functions, installed from a pinned PerspectiveGap commit.
- Publishes generated benchmark meta/docs for both English and Chinese docs.
- Adds a mock-LLM EvalScope path test that loads the actual ModelScope dataset and runs scoring.

# Local validation

```bash
.venv/bin/python -m pytest tests/benchmark/test_perspective_gap.py -v -s -p no:warnings
```

Result: `8 passed in 1.35s`.

```bash
.venv/bin/python -m pytest tests/benchmark/test_eval.py::TestNativeBenchmark::test_perspective_gap_mock -v -s -p no:warnings
```

Result: `1 passed in 4.54s`.

```bash
PATH=.venv/bin:$PATH make lint
```

Result: flake8, isort, yapf, pre-commit hooks passed.

```bash
PATH=.venv/bin:$PATH make docs-pipeline BENCHMARK="perspective_gap_role_assignment perspective_gap_prompt_writing" FORCE=1
PATH=.venv/bin:$PATH make docs-generate
```

Result: stats/meta generated for both tasks with 220 samples each; `docs-pipeline` translation step could not run locally because `DASHSCOPE_API_KEY` was unavailable, then docs were generated from the updated meta files.

# Dataset verification

Verified the ModelScope `evalscope/PerspectiveGap` `test` split exposes the fields used by the adapter: `reference_need_sets`, `distractor_id`, `fragments`, `role_assignment_prompt`, `prompt_writing_prompt`, `evaluation_id`, `scenario_id`, `shuffle_seed`, and `roles`.


Yunnglin (commented):
Thanks for putting this together — the lazy import for the optional scorer and reusing the official `perspective_gap.scoring` functions instead of reimplementing the metrics are both nice touches. Before this can be merged though, there are a few things to sort out:

1. We've mirrored the dataset to ModelScope at `evalscope/PerspectiveGap`, so please switch `DATASET_ID` in the adapter from the HF `sun1245/PerspectiveGap` to `evalscope/PerspectiveGap` and drop the `dataset_hub='huggingface'` requirement from the docs/examples. Also worth double-checking the mirrored data still has the same fields your `record_to_sample` relies on (`reference_need_sets`, `distractor_id`, etc.) — just to be safe after the move.

2. CI's lint check is currently red — `isort`, `yapf` and `double-quote-string-fixer` all flagged changes. The double-quote one is on `INSTALL_HINT` in the adapter (should be single quotes here), and the import order in the test file needs an isort pass. Running `make lint` locally and committing the auto-fixes should clear this up.

3. The docs page you added by hand (`docs/en/benchmarks/perspective_gap.md`) actually needs to come from the doc pipeline, not be written manually — that's how every other benchmark's docs/meta get generated in this repo. Could you delete it and instead run:
   ```bash
   make docs-pipeline BENCHMARK="perspective_gap_role_assignment perspective_gap_prompt_writing" FORCE=1
   make docs-generate
   ```
   That'll also give you the `_meta/*.json` files and the Chinese translation, both of which are missing right now.

4. On the tests — right now `test_perspective_gap.py` mocks `perspective_gap.scoring` via `sys.modules` and only checks `record_to_sample`/`match_score` in isolation. It never actually goes through `run_task`, and it's also not wired into any CI job (`CI Tests Lite` only runs `test_ci_lite`), so honestly these tests aren't being run anywhere right now. Could you add something like the other benchmarks in `tests/benchmark/test_eval.py`:
   ```python
   def test_perspective_gap_mock(self):
       self._run_dataset_test('perspective_gap_role_assignment', limit=5, use_mock=True)
       self._run_dataset_test('perspective_gap_prompt_writing', limit=5, use_mock=True)
   ```
   That covers the real data loading + MockLLM + scoring path end to end, and actually runs in CI. Your existing unit tests are still nice to keep around for the scoring-dispatch edge cases.

5. Also noticed the "Local validation" section says "Result: 5 passed", but the current test file actually has 8 test functions — looks like that number is from an earlier commit and wasn't updated after the later "Harden PerspectiveGap scoring adapter" commit added more tests. Could you re-run and paste the actual output (not just a summary line)? Since these tests mock out the scorer entirely, it'd also help to see a real end-to-end run (even a mock-LLM one) against the actual dataset once it's pointed at `evalscope/PerspectiveGap`.

6. Small one — the `perspective-gap` package in `requirements.txt` is pulled straight from the repo's HEAD via git+https, with no pinned commit/tag. Might be worth pinning it so results don't silently change if the upstream repo gets updated.

7. Nitpick, not blocking: `net_match_score` / `required_coverage` / `boundary_precision` / `distractor_leakage` only show up in `score.metadata` right now — might be nice to have them as extra entries in `metric_list` so they show up in the report too, but not a must.

Yunnglin (changes requested):

WhymustIhaveaname:
Thanks for the detailed review. I pushed an update addressing the requested changes.

Main changes:

- Switched the default dataset to `evalscope/PerspectiveGap`.
- Checked the mirrored `test` split fields used by the adapter.
- Replaced the hand-written docs with generated en/zh benchmark docs and `_meta` files.
- Added a CI-lite mock test in `tests/benchmark/test_eval.py`.
- Pinned the `perspective-gap` dependency to a fixed commit.
- Added the PerspectiveGap submetrics to `metric_list`.
- Ran `make lint` and committed the pre-commit fixes.

Validation:

- focused PerspectiveGap tests: 8 passed
- CI-lite mock test: 1 passed
- `make lint`: passed

One note: `make docs-pipeline` generated stats/meta successfully, but the translation step needs `DASHSCOPE_API_KEY`, which I do not have locally. I filled the generated `readme.zh` entries and then ran `make docs-generate` successfully.

Please let me know if anything else should be adjusted.

Yunnglin (approved):
LGTM

WhymustIhaveaname:
Thanks for the review and fix!

PR #1463, opened 2026-07-08
fix(openai): include delta.reasoning in streaming TTFT

git-jxj:
## Summary

This PR fixes streaming TTFT and TPOT metrics for OpenAI-compatible providers, such as GLM-5.2 deployments, that return reasoning tokens in `choice.delta.reasoning` instead of `choice.delta.reasoning_content`.

The change also preserves `delta.reasoning` in the aggregated assistant message. Both synchronous and asynchronous stream collectors are covered.

## Root Cause

The current first-token detection only recognizes non-empty values from:

- `delta.content`
- `delta.reasoning_content`
- `delta.tool_calls`

Some OpenAI-compatible providers stream their reasoning tokens through `delta.reasoning`. Those chunks are ignored until a later `delta.content` chunk arrives.

This records TTFT too late. EvalScope calculates TPOT as:

```text
(latency - ttft) / (output_tokens - 1)
```

The provider's usage metadata still includes the reasoning tokens in `output_tokens`. A late TTFT therefore makes the decoding interval too short while leaving the token count unchanged, significantly underestimating TPOT and overestimating decoding performance.

## Changes

- Prefer the standard `delta.reasoning_content` field.
- Fall back to `delta.reasoning` when `reasoning_content` is absent.
- Treat either reasoning field as meaningful first-token content for TTFT.
- Preserve the selected reasoning field in the aggregated assistant message.
- Apply identical behavior to synchronous and asynchronous stream aggregation.
- Add deterministic regression tests for reasoning aggregation and TTFT.

The fallback avoids concatenating both fields if a provider returns both aliases in the same chunk.

## Reproduction

The regression tests simulate this stream:

```text
10.25s: delta.reasoning = "thinking"
11.00s: delta.content = "answer"
```

On the unmodified `main` branch, both stream collectors return:

```text
(reasoning_content, ttft) = (None, 1.0)
```

The expected result is:

```text
(reasoning_content, ttft) = ("thinking", 0.25)
content = "answer"
```

## Validation

```bash
python -m pytest \
  tests/models/test_openai_stream_reasoning.py \
  tests/models/test_openai_reasoning.py -q
```

Result:

```text
14 passed
```

The new tests fail on the unmodified `main` branch for both synchronous and asynchronous collection with:

```text
assert (None, 1.0) == ("thinking", 0.25)
```

### GLM-5.2 provider validation

**Before**

GLM-5.2 performance metrics before the fix

**After**

GLM-5.2 performance metrics after the fix

After applying the change, a 30-sample AIME26 performance run using a GLM-5.2 deployment reported:

```text
Avg latency:       387.963 s
Avg TTFT:          925.32 ms
Avg TPOT:          17.75 ms
Avg output tokens: 21769.4
```

The reasoning stream is now recognized as the start of model output, and the resulting TTFT/TPOT values are consistent with the observed end-to-end latency and output-token volume.

## Scope

This PR only changes OpenAI-compatible stream aggregation and the timing derived from the first meaningful stream chunk. It does not change non-streaming responses, request parameters, provider usage metadata, tool-call aggregation, or retry behavior.

### DIFF
--- a/evalscope/models/utils/openai.py
+++ b/evalscope/models/utils/openai.py
@@ -823,13 +823,13 @@ def collect_stream_response(
         collected_chunks.append(chunk)
         for choice in chunk.choices:
             reasoning_content = getattr(choice.delta, 'reasoning_content', None)
-            if reasoning_content is None:
+            if reasoning_content is None or reasoning_content == '':
                 reasoning_content = getattr(choice.delta, 'reasoning', None)
 
             # Detect first meaningful content chunk for TTFT
-            has_content = ((choice.delta.content is not None and choice.delta.content != '') or (
-                reasoning_content is not None and reasoning_content != ''
-            ) or (hasattr(choice.delta, 'tool_calls') and choice.delta.tool_calls))
+            has_content = ((choice.delta.content is not None and choice.delta.content != '')
+                           or (reasoning_content is not None and reasoning_content != '')
+                           or (hasattr(choice.delta, 'tool_calls') and choice.delta.tool_calls))
             if ttft is None and has_content:
                 ttft = time.monotonic() - t_start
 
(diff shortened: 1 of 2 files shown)


Yunnglin:
Thanks for the fix. The functional tests around streaming reasoning look good locally.\n\nCould you please run `make lint` and fix any lint / pre-commit issues before this gets merged? Once lint is clean, this should be good from my side.

git-jxj:
> Thanks for the fix. The functional tests around streaming reasoning look good locally.\n\nCould you please run `make lint` and fix any lint / pre-commit issues before this gets merged? Once lint is clean, this should be good from my side.

thanks for the review. I've run `make lint` and it's clean now. Also addressed the Copilot comments by treating empty `reasoning_content` as absent and falling back  to `delta.reasoning`. Added regression tests for that case.

Yunnglin (approved):

PR #1464, opened 2026-07-08
fix(openai): retry interrupted streaming responses

git-jxj:
## Summary

This PR retries OpenAI-compatible streaming requests when the response stream is interrupted after the initial request has already been established.

The fix applies to both synchronous and asynchronous generation. Partial output from a failed attempt is discarded, and the full model request is retried using the existing `retries` and `retry_interval` configuration.

## Root Cause

The current retry boundary only wraps:

```python
chat.completions.create(...)
```

For a streaming response, `create()` returns a stream object before all response chunks have been received. The stream is consumed later by `collect_stream_response()` or `async_collect_stream_response()`, outside the retry boundary.

If an upstream gateway closes the SSE connection after emitting partial output, stream consumption raises an exception such as a connection or protocol error. Although the request is configured with retries, the interrupted model call fails immediately because the exception occurs after `retry_call()` has returned.

With `ignore_errors=True`, evaluation may continue while the failed sample is missing from predictions, reviews, and the final report.

## Changes

- Treat request creation and complete stream consumption as one retryable operation.
- Retry the full request when synchronous stream consumption fails.
- Retry the full request when asynchronous stream consumption fails.
- Discard chunks collected by a failed attempt instead of mixing them into the successful retry.
- Keep non-streaming completion handling within the same operation without changing its response behavior.
- Add deterministic regression tests for synchronous and asynchronous interrupted streams.

## Reproduction

The tests simulate an upstream stream that behaves as follows:

```text
attempt 1: emit "discarded partial response" -> raise ConnectionError
attempt 2: emit "complete response" -> finish normally
```

On the unmodified `main` branch, both tests fail at the first interrupted stream:

```text
ConnectionError: stream interrupted by upstream gateway
```

The second request is never attempted even though `retries=2`.

After this change:

```text
attempts == 2
result.content == "complete response"
```

This also verifies that partial content from the failed attempt is not included in the final response.

## Validation

```bash
```

Result:

```text
14 passed
```

Additional checks:

```bash
python -m py_compile \
  evalscope/models/openai_compatible.py \
  tests/models/test_openai_stream_retry.py

git diff --check
```

## Agent and Tool-call Safety

The stream is fully consumed before the model output is added to message history or any returned tool call is executed. Retrying an interrupted stream therefore does not repeat an already executed sandbox command or tool call.

## Scope

This PR does not change retry counts, retry intervals, request parameters, non-streaming response schemas, stream aggregation rules, or error handling after all configured attempts are exhausted.

### DIFF
--- a/evalscope/models/openai_compatible.py
+++ b/evalscope/models/openai_compatible.py
@@ -145,16 +145,21 @@ class OpenAICompatibleAPI(ModelAPI):
             t_start = time.monotonic()
             ttft: Optional[float] = None
 
-            # generate completion and save response for model call
-            completion = retry_call(
-                self.client.chat.completions.create,
+            # A streaming request is not complete when create() returns: the
+            # connection may still fail while its chunks are being consumed.
+            # Retry the whole request so a partial response is discarded and
+            # replaced by one complete response.
+            def _create_and_collect() -> Tuple[ChatCompletion, Optional[float]]:
+                raw_completion = self.client.chat.completions.create(**request)
+                if isinstance(raw_completion, ChatCompletion):
+                    return raw_completion, None
+                return collect_stream_response(raw_completion, request_start=t_start)
+
+            completion, ttft = retry_call(
+                _create_and_collect,
                 retries=config.retries,
                 sleep_interval=config.retry_interval,
-                **request
             )
-            # handle streaming response
-            if not isinstance(completion, ChatCompletion):
-                completion, ttft = collect_stream_response(completion, request_start=t_start)
 
             total_time = time.monotonic() - t_start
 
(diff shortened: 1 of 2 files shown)


Yunnglin:
Thanks for the fix. The streaming retry behavior looks good in local regression testing.\n\nCould you please run `make lint` and fix any lint / pre-commit issues before this gets merged? Once lint is clean, this should be good from my side.

git-jxj:
> Thanks for the fix. The streaming retry behavior looks good in local regression testing.\n\nCould you please run `make lint` and fix any lint / pre-commit issues before this gets merged? Once lint is clean, this should be good from my side.

thanks for the review. I've run `make lint` and it's clean. Also added type hints to the test helpers as suggested by Copilot. PTAL.

Yunnglin (approved):

PR #1488, opened 2026-07-14
fix: support code execution scoring on Windows

### LINKED ISSUE #1486: HumanEval / LiveCodeBench local code-execution scoring silently returns 0 on Windows
On Windows, time_limit() in evalscope/benchmarks/humaneval/utils.py calls signal.setitimer(signal.ITIMER_REAL, ...) / signal.SIGALRM, which don't exist on Windows. Entering the context raises AttributeError, which is caught by the except BaseException around exec() and recorded as failed: ..., so every sample scores as failed and pass@k aggregates to 0 with no error surfaced. The outer process-level timeout (p.join(timeout+1) + p.kill()) already enforces the time limit, so the inner signal-based timeout is redundant on Windows and can be platform-guarded without losing the limit. evalscope/benchmarks/live_code_bench/testing_util.py has the same pattern. Reproduced on Windows 11. Happy to open a PR with a platform guard + a small test.

toffee-desuwa:
Fixes #1486.

### Summary
- HumanEval: skip the inner POSIX interval timer when `signal.setitimer` is unavailable
- LiveCodeBench: route all `signal.alarm()` set/clear points through a capability-guarded helper, and guard the `SIGALRM` handler registration
- add regression tests for the HumanEval path and for both LiveCodeBench paths (`call_based` and `standard_input`) on platforms without POSIX signal timers

Windows does not provide `setitimer`, `SIGALRM`, or `alarm`, so the previous inner timeout setup raised `AttributeError` before a known-correct completion could execute; that exception was swallowed and every sample was scored as failed. This guards the inner signal-based timeout by capability while leaving POSIX behavior unchanged.

The existing parent-process safeguards still bound runtime on all platforms: HumanEval uses `join(timeout + 1)` then `kill()`, and LiveCodeBench applies its global process timeout (`codegen_check_correctness`). POSIX platforms keep their existing inner signal timeouts byte-for-byte.

Per @Yunnglin's note, this keeps the sandbox path as the recommended default and only hardens the native local-execution path; it covers both the HumanEval `setitimer`/`SIGALRM` path and all LiveCodeBench `signal.alarm()` usages.

### Test plan
- `python -m pytest tests/benchmark/test_code_execution_timeouts.py -vv` — 3 passed
- ran known-correct HumanEval and LiveCodeBench (call_based + standard_input) samples on Windows 11; all scored correctly

### DIFF
--- a/evalscope/benchmarks/humaneval/utils.py
+++ b/evalscope/benchmarks/humaneval/utils.py
@@ -88,7 +88,7 @@ def check_correctness(problem: Dict, completion: str, timeout: float, completion
 @contextlib.contextmanager
 def time_limit(seconds: float):
 
-    if not (hasattr(signal, 'setitimer') and hasattr(signal, 'SIGALRM')):
+    if not (hasattr(signal, 'setitimer') and hasattr(signal, 'SIGALRM') and hasattr(signal, 'ITIMER_REAL')):
         yield
         return
 
--- a/evalscope/benchmarks/humaneval/utils.py
+++ b/evalscope/benchmarks/humaneval/utils.py
@@ -88,6 +88,10 @@ def check_correctness(problem: Dict, completion: str, timeout: float, completion
 @contextlib.contextmanager
 def time_limit(seconds: float):
 
+    if not hasattr(signal, 'setitimer'):
+        yield
+        return
+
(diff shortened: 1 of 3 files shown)


Yunnglin (approved):
LGTM

PR #1498, opened 2026-07-20
fix(perf): add /v1/rerank endpoint support

hubutui:
## Summary
- add /v1/rerank to the known OpenAI-compatible perf endpoints
- keep existing /v1/reranks support unchanged

## Tests
- Not run; one-line endpoint allowlist change.


Yunnglin (approved):
LGTM

PR #1499, opened 2026-07-21
fix(perf): exclude non-stream requests from TTFT/TPOT/ITL metrics

qiumuyang:
## Summary

When a benchmark run mixes `stream: true` and `stream: false` requests, streaming-specific metrics (TTFT / TPOT / ITL) are now computed over **stream requests only**. Pure-stream and pure-non-stream runs are unchanged.

## Motivation

For non-stream requests, `first_chunk_latency = query_latency` (entire request time) and `time_per_output_token = 0`. When mixed with real stream requests this **inflates TTFT** and **deflates TPOT**.

Mixed runs are reachable via `workload_trace` (per-request `stream` from original trace) and `line_by_line` (JSONL body `stream` field overrides CLI flag).

### Example

8 streaming (TTFT ~50 ms) + 2 non-streaming (latency ~2.5 s):

| Metric   | Before                             | After (stream only) |
|----------|------------------------------------|---------------------|
| Avg TTFT | `(8×0.05 + 2×2.5) / 10 = 0.54 s`  | `0.05 s`            |
| Avg TPOT | deflated by two `0` contributions  | stream-only mean    |

## Changes

| File | What changed |
|------|--------------|
| `benchmark_util.py` | Add `is_stream` flag and `is_stream_body()` helper; stream-only accumulators for TTFT/TPOT/ITL with non-stream fallback; expose stream/non-stream counts. |
| `default_api.py` | Set `is_stream = True` in SSE branch; body-based fallback on failure. |
| `openai_responses_api.py` | Set `is_stream = True` in SSE branch; body-based fallback on failure. |
| `http_client.py` | Body-based `is_stream` on timeout/connection-error paths. |
| `db_util.py` | Persist `is_stream` column; percentile TTFT/TPOT/ITL/decode use stream rows only (fallback for pure non-stream). |
| `perf_constants.py` | `STREAM_REQUESTS` / `NON_STREAM_REQUESTS` constants. |
| `perf_models.py` | Count fields in `BenchmarkSummary`; conditional display when mixed. |

## Testing

```
$ pytest tests/perf/test_stream_metrics.py -q
9 passed
```

### DIFF
--- a/evalscope/perf/core/http_client.py
+++ b/evalscope/perf/core/http_client.py
@@ -4,7 +4,7 @@ import time
 from typing import TYPE_CHECKING
 
 from evalscope.perf.arguments import Arguments
-from evalscope.perf.utils.benchmark_util import BenchmarkData
+from evalscope.perf.utils.benchmark_util import BenchmarkData, is_stream_body
 from evalscope.utils.argument_utils import SECRET_HEADER_KEYS, get_secret_value
 from evalscope.utils.logger import get_logger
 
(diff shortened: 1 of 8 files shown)


Yunnglin (approved):
LGTM. Bucketing TTFT/TPOT by the actually-streamed response is the right fix, and the pure-non-stream fallback keeps existing runs unchanged.

Pushed two small refinements to the branch (`e1517dad`):
1. Dropped the redundant ITL stream bucket — non-stream requests contribute an empty ITL list, so it's already stream-only by construction (TTFT/TPOT/decode correctly stay on the stream subset).
2. Hardened `is_stream_body` against non-dict bodies on error/timeout paths.

Tests (9 passed) + `make lint` clean. Thanks!

PR #1503, opened 2026-07-22
fix(perf): generate HTML report for open-loop runs with no fixed rate (rate=-1.0)

qiumuyang:
## Problem

Open-loop perf runs that don't pin a fixed request rate produce **no `perf_report.html`** — it's silently skipped, even though `performance_summary.txt`, `benchmark_summary.json`, `benchmark_percentile.json`, and `benchmark_data.db` are all written correctly.

For example, `workload_trace` replay (driven by trace arrival offsets, no `--rate`) ends the run with:

```
INFO: Performance summary saved to: .../performance_summary.txt
WARNING: No benchmark runs found under .../outputs/<name>. Skipping HTML report.
```

## Root cause

When rate is unset, the benchmark records `Request Rate = -1.0` (sentinel) and writes the run sub-directory as `rate_-1.0_number_<M>`. The loader regex has no sign in its character class, so the directory never matches and `RunLoader.load_all()` drops it:

```python
# evalscope/perf/utils/report/perf_data.py
_RATE_RE = re.compile(r'^rate_([\d.]+)_number_(\d+)$')   # rejects the leading '-'
```

With zero runs returned, `gen_perf_html_report()` logs *"No benchmark runs found ... Skipping HTML report"* and returns `''`.

## Fix

Allow an optional leading sign on the rate capture group:

```python
_RATE_RE = re.compile(r'^rate_(-?[\d.]+)_number_(\d+)$')
```

Positive-rate run dirs (`rate_1.5_number_4`) still match; `parallel_*` is unaffected.

## Verification

Re-ran a `workload_trace` open-loop run (rate=-1.0, 30 reqs) against a vLLM endpoint (`--enable-prompt-tokens-details`):

- **Before:** `WARNING: No benchmark runs found ... Skipping HTML report.` — no `perf_report.html`.
- **After:** `HTML report generated: .../perf_report.html` (165 KB interactive report, 26/30 success; 4 failures are unrelated `ServerDisconnectedError`).

Run dir confirmed named `rate_-1.0_number_30` — i.e. the negative-rate path that previously failed discovery.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/evalscope/perf/utils/perf_models.py
+++ b/evalscope/perf/utils/perf_models.py
@@ -434,14 +434,15 @@ class RunData(BaseModel):
     def name(self) -> str:
         """Human-readable run label.
 
-        * ``parallel_<N>_number_<M>`` directories  → "Parallel N / Number M"
-        * ``rate_<R>_number_<M>`` directories       → "Rate R rps / Number M"
+        * closed-loop (``parallel_<N>_number_<M>``) → "Parallel N / Number M"
+        * open-loop with a fixed rate               → "Rate R rps / Number M"
+        * open-loop with no fixed rate (rate=-1)    → "Open-loop / Number M"
         """
-        import re
-        m = re.match(r'^rate_([\d.]+)_number_(\d+)$', self.dir_name)
-        if m:
-            return f'Rate {m.group(1)} rps / Number {m.group(2)}'
-        return f'Parallel {self.parallel} / Number {self.number}'
+        if not self.is_open_loop:
+            return f'Parallel {self.parallel} / Number {self.number}'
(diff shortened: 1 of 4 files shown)


Yunnglin:
Reproduced and verified. Pushed a small follow-up (70ce1c5c): derive open-loop from `RunData.is_open_loop` so the run is labeled "Open-loop / Number M" with a "Rate (req/s)" axis (instead of "Parallel 0" / "Concurrency"), plus a unit test. lint + perf tests pass. LGTM!

Yunnglin (approved):
LGTM

PR #1510, opened 2026-07-23
fix(web): accept completed invoke status

afox666:
## Summary

  Allow the Web dashboard to accept `completed` as a valid invoke response status.

## Problem

image

  The evaluation and performance invoke endpoints return the following status after a successful task:

  ```json
  {
    "status": "completed",
    "task_id": "..."
  }
```
  However, the frontend runtime schema only accepted:

  - ok
  - error
  - stopped

  As a result, successfully completed tasks were rejected by Zod and displayed as:

  DomainError: Response validation failed

## Changes

  - Add completed to invokeStatusSchema.
  - Preserve ok for backward compatibility.
  - Update the endpoint schema fixture to reflect the current backend response.
  - Add regression tests covering all supported statuses.
  - Add a test ensuring unknown statuses are still rejected.

## Testing

  - 37 endpoint schema tests passed.
  - TypeScript and Vite production build passed.
  - ESLint passed for the modified files.

## Compatibility

  This change is backward compatible. Existing ok, error, and stopped responses remain supported.

### DIFF
--- a/evalscope/web/src/api/schemas/endpointSchemas.test.ts
+++ b/evalscope/web/src/api/schemas/endpointSchemas.test.ts
@@ -29,7 +29,7 @@ const ENDPOINT_CASES: Array<{ name: string; schema: ZodType; valid: unknown }> =
   },
   { name: 'log', schema: logResponseSchema, valid: { text: '', head_line: 0, tail_line: 0, total_lines: 0 } },
   { name: 'task status', schema: taskStatusResponseSchema, valid: { status: 'running', task_id: 'task-1' } },
-  { name: 'eval invoke', schema: evalInvokeResponseSchema, valid: { status: 'ok', task_id: 'task-1' } },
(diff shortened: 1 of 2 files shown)


Yunnglin (approved):
LGTM ✅

Verified locally:
- Reproduced the bug: with the old schema (`ok/error/stopped`), a `completed` status fails Zod validation → `DomainError: Response validation failed`. Confirmed via vitest (the `completed` case failed, 1/37).
- Confirmed the fix: after adding `completed`, all 37 endpoint-schema tests pass.
- Backend contract check: eval/perf `/invoke` only ever return `completed` / `error` / `stopped`, and the shared `evalInvokeResponseSchema` now covers all three, fixing both eval and perf paths in one change.
- Ran eval-service endpoint tests (Flask test client, subprocess mocked): 8/8 pass, including the success path returning `status: 'completed'`.

Minimal, strict (unknown statuses still rejected), and backward compatible. Thanks!

PR #1536, opened 2026-08-03
fix: assign result of str.replace() in math parser

LeoYueDev:
string.replace() returns a new string and does not mutate in place. The return values were discarded, so quotes were never actually removed from answer strings, causing math_equal to report false negatives.


Yunnglin (approved):
LGTM

PR #1539, opened 2026-08-04
fix(ifbench): download correct NLTK English tagger

git-jxj:
## Summary

This PR fixes the NLTK resource setup used by IFBench for English part-of-speech tagging.

IFBench requests `averaged_perceptron_tagger_eng`, but the resource helper currently checks and downloads the legacy `averaged_perceptron_tagger` package. Setup can therefore appear to succeed while the resource required by current NLTK remains unavailable.

## Root Cause

The mirror entry for `averaged_perceptron_tagger_eng` maps the requested resource to different legacy names:

```text
requested:  taggers/averaged_perceptron_tagger_eng/
checked:    taggers/averaged_perceptron_tagger
downloaded: averaged_perceptron_tagger.zip
```

Current NLTK loads the English perceptron tagger from `taggers/averaged_perceptron_tagger_eng/`, so downloading the legacy package does not satisfy the lookup.

The helper also catches setup failures and only logs them, which allows evaluation to continue until NLTK raises a less direct lookup error later.

## Changes

- Check and download `averaged_perceptron_tagger_eng.zip` from the official NLTK data repository.
- Verify the resource is discoverable after extraction.
- Propagate download and post-download validation failures to the caller.
- Keep existing resources as a no-op and preserve the existing cache behavior.
- Add regression tests for existing, successfully downloaded, still-missing, and failed-download resources.

## Validation

```bash
python -m pytest tests/test_resource_utils.py -q
```

Result:

```text
4 passed
```

The relevant pre-commit hooks also passed:

```text
flake8, isort, yapf, trailing whitespace, YAML, end-of-file,
```

The configured NLTK 3.10.0 runtime resolves the corrected resource path successfully, and the official `_eng` archive URL returns HTTP 200.

## Scope

This change only affects automatic setup for mapped NLTK resources. It does not change IFBench instructions, scoring rules, dataset loading, or already-installed resources.

### DIFF
--- a/evalscope/utils/resource_utils.py
+++ b/evalscope/utils/resource_utils.py
@@ -27,6 +27,7 @@ def check_nltk_data(name: str) -> None:
     import nltk
 
     GITEE_MIRROR = 'https://gitee.com/yzy0612/nltk_data/raw/gh-pages/packages'
+    NLTK_DATA_RAW = 'https://raw.githubusercontent.com/nltk/nltk_data/gh-pages/packages'
 
     MIRROR_MAP = {
         'punkt_tab': {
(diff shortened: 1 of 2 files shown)


Yunnglin (approved):
LGTM

PR #1547, opened 2026-08-05
feat(perf): add startup_ramp_seconds for multi-turn benchmark

yonlunwu:
Staggers worker spawn across the first `parallel` slots using a Poisson schedule (rescaled so worker n-1 lands exactly at the configured duration). Once the ramp completes, `--rate` resumes governing turn-to-turn pacing.

Applies at the start of both warmup and benchmark phases. Two threshold fields gate the ramp against configurations where the curve would not be visible: `warmup_ramp_min_conversations` and
`benchmark_ramp_min_parallel` (each default 3, set 0 to disable).

### DIFF
--- a/evalscope/perf/arguments.py
+++ b/evalscope/perf/arguments.py
@@ -200,6 +200,26 @@ class Arguments(BaseArgument):
     in_flight_task_multiplier: int = 2
     """Max scheduled tasks = parallel * this multiplier."""
 
+    startup_ramp_seconds: Optional[float] = None
+    """Stagger multi-turn workers' initial spawn across this many seconds
+    via a Poisson schedule; worker 0 fires at phase start, worker n-1 at
+    exactly ``value`` seconds after. Disabled when None / 0. Applies
+    once at each phase start; takes no part in steady-state operation
+    (``--rate`` still governs turn-to-turn pacing). Auto-skipped on
+    warmup when ``warmup_count <= warmup_ramp_min_conversations`` and on
+    benchmark when ``parallel <= benchmark_ramp_min_parallel``.
+    Single-turn strategies ignore this.
+
+    """
+
+    warmup_ramp_min_conversations: int = 3
+    """Skip the warmup-phase startup ramp when ``warmup_count <=`` this
+    value. Set 0 to disable the guardrail."""
+
+    benchmark_ramp_min_parallel: int = 3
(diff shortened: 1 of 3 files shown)


yonlunwu:
@Yunnglin In the current multi-round dialogue evaluation, after the warm-up phase is completed, there will be 'parallel' workers simultaneously sending requests, which is equivalent to a cold concurrent operation. This will impose a significant pressure on the cluster. This parameter is a meaningful feature in actual testing. It achieves an initial ramp-up and flow addition, providing a relatively good buffer for the system.

I sincerely hope that you could please take a look at it and provide a review. Thank you very much!

Yunnglin:
Thanks for the PR and for digging into the phase-start burst!

We're currently evaluating the underlying problem and discussing whether a different approach fits better — in particular, aligning the multi-turn warmup semantics with the upstream vLLM multi-turn benchmark (warmup sends only the first turn of each conversation and writes back the real reply, so the benchmark phase starts from turn 2 and the cold prefill stays outside the measured window).

Holding off on merging until we settle on the direction. Will follow up here with a conclusion.

Yunnglin:
Thanks for this, and sorry for the long silence.

**Your diagnosis was right** — the phase boundary drains occupancy to zero, so the next phase releases `parallel` requests at once against an idle server. On the single-turn path that overstated the reported `p99` by 4.19x (1609 ms vs 384 ms steady-state). Fixed in #1641, which **did not touch multi-turn** — so what you reported here is still unfixed, and #1648 tracks it.

We did conclude the ramp isn't the right shape, so I'm closing this PR:

- It's applied to the benchmark phase, so it trades a burst at the start of the measured window for an under-loaded stretch *inside* it — as your help text says, "the ramp window consumes part of any `--duration` budget". #1641 instead removes the barrier, so each warmup completion hands its slot straight to a measured request and the window opens saturated: no new flags, and no unrepresentative stretch anywhere in the run.
- `_spawn_workers` returns `workers` only at the end, so a cancellation mid-ramp skips the `return` and the caller's `finally` has nothing to cancel — the tasks already created leak. (From reading the diff, not a repro.)

#1648 proposes the same barrier removal for multi-turn plus upstream's first-turn warmup, which also fixes something the ramp doesn't touch: warmup and benchmark currently pull *disjoint* conversations, so every measured conversation still pays a cold first turn. Happy to have you take it on — and thanks for surfacing this.

PR #1554, opened 2026-08-07
fix: validate service output directory

heliubj18:
## Background

When running the example from the [Quick Experience](https://evalscope.readthedocs.io/en/latest/get_started/visualization.html#quick-experience) guide:

```bash
evalscope service --outputs examples/viz
```

`examples/viz` may not exist in the local working directory. The backend service still starts, and the frontend then shows the confusing error:

```text
root_path is required and must be an existing directory
```

## Summary

- Validate --outputs as an existing directory during argparse processing.
- Prevent the service from starting with an invalid output root.
- Replace the misleading frontend error with a clear CLI error.

## Before

An invalid --outputs path allowed the backend service to start, then the frontend reported:
`root_path is required and must be an existing directory`.

## After

```text
usage: EvalScope Command Line tool service ...
EvalScope Command Line tool service: error: argument --outputs: output directory does not exist: /private/tmp/not_exist_dir
```

The backend now fails during argument parsing and exits with status code 2. The service does not start, so the frontend is not opened with an invalid output root.

## Tests

- `pytest -q tests/cli/test_service.py`
- CLI invalid-output smoke test exits with code 2

### DIFF
--- a/evalscope/cli/start_service.py
+++ b/evalscope/cli/start_service.py
@@ -1,6 +1,7 @@
 # Copyright (c) Alibaba, Inc. and its affiliates.
 """CLI command for starting the EvalScope Flask service."""
-from argparse import ArgumentParser
+import os
+from argparse import ArgumentParser, ArgumentTypeError
 
 from evalscope.cli.base import CLICommand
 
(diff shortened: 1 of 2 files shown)


Yunnglin:

pip install pre-commit
pre-commit install
pre-commit run --all-files

heliubj18:
> Thank you for your PR, could you fix the lint issues? Run the following command:
> 
> pip install pre-commit pre-commit install pre-commit run --all-files

Done，thanks！

Yunnglin (approved):
LGTM

PR #1563, opened 2026-08-11
fix(web): allow null reasoning_tokens in report content blocks

Dhru1001:
ContentReasoning.reasoning_tokens is Optional[int] = None on the Python side, which serializes as an explicit JSON null rather than an absent key. The zod schema only allowed number|undefined, so any reasoning model that doesn't report a token count (e.g. no
completion_tokens_details.reasoning_tokens from the API) failed report validation on every subset - reproduced against a real report (eval_1786353041334) and confirmed fixed.

Widen ReasoningBlock's tokens prop to number | null to match; it already guards with `tokens != null` at render time.

### DIFF
--- a/evalscope/web/src/api/schemas/reports.schema.test.ts
+++ b/evalscope/web/src/api/schemas/reports.schema.test.ts
@@ -64,4 +64,19 @@ describe('loadReportResponseSchema real report compatibility', () => {
 
     expect(result.success).toBe(true)
   })
+
+  it('accepts a reasoning block with a null reasoning_tokens, as emitted when the backend has no token count', () => {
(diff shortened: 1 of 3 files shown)


Dhru1001:
@Yunnglin

Yunnglin (approved):
LGTM. Verified against the Python-side contract: `ContentReasoning.reasoning_tokens` is `Optional[int] = None`, which Pydantic serializes as an explicit JSON `null`, so the previous `z.number().optional()` rejected valid reasoning blocks and failed report validation for every subset. Widening to `.nullable().optional()` matches the backend contract, and `ReasoningBlock` already guards with `tokens != null`. CI green, focused change with a regression test. Thanks @Dhru1001!

PR #1569, opened 2026-08-12
feat(benchmarks): add native Indic-language benchmark adapters (milu,…

Dhru1001:
Add 5 evalscope-native benchmark adapters that can now run directly through EvalScope’s built-in vLLM/OpenAI-API serving path 
### Added benchmarks

* `milu` — `ai4bharat/MILU` (11 languages, gated dataset, requires `HF_TOKEN`)
* `arc_indic` — `sarvamai/arc-challenge-indic` (11 languages)
* `triviaqa_indic` — `sarvamai/trivia-qa-indic-mcq` (11 languages)
* `indic_boolq` — `sarvamai/boolq-indic` (11 languages)
* `gsm8k_indic` — `sarvamai/gsm8k-indic` (21 language/script variants)

### Validation

Live-verified the following adapters end-to-end against their real HuggingFace datasets:

* `arc_indic`
* `triviaqa_indic`
* `indic_boolq`
* `gsm8k_indic`

Validation covered:

* dataset loading
* prompt formatting
* answer extraction
* end-to-end evaluation execution

Enables native Indic-language evaluation inside EvalScope and improves compatibility with existing vLLM and OpenAI-compatible serving workflows.


Dhru1001:
@Yunnglin

Dhru1001:
@Yunnglin I’m thinking of maintaining a single branch for the additional benchmarks and features I’m planning to add, while keeping it regularly synced with main. Since there are multiple benchmarks/features to add, maintaining separate branches for each would add a fair amount of branch management and merge/rebase overhead. A single feature branch would make it easier to manage the work while still keeping it in sync with main. 

If you need to maintain different feature branch i fine with that also I was just looking to speed up things, while making my opensource contributions as well. 

For sure if would love review's on code. As Might loose out minute details. 

Let me know your views on the same

Yunnglin (commented):
Thanks for this @Dhru1001 — solid work. I verified the five adapters end-to-end before reviewing, and the data mapping is correct in all of them. Below are the review findings first, then my answer on the branching question.

## Verification I ran

All five adapters run end-to-end through the default hub path (`mock_llm`, `limit=2`, prompts render correctly for Devanagari/Latin input):

```
PASS milu | arc_indic | triviaqa_indic | indic_boolq | gsm8k_indic
```

Field-mapping integrity check against the full splits — no issues found:

| Benchmark | Check | Result |
| --- | --- | --- |
| `milu` | `target` ∈ `option1..4`, no empty options | Hindi 14831 / English 13535 / Tamil 6372, 0 anomalies |
| `arc_indic` | `answerKey` values, choice count | all A–D, `n_choices` always 4 |
| `triviaqa_indic` | `answer` index in range | 17944 rows/lang, 0 out-of-range (0-based confirmed) |
| `indic_boolq` | `label` vs `answer` polarity | 35970 rows, 0 mismatches; 3270 per language × 11 |
| `gsm8k_indic` | `####` delimiter present | 1319 rows × (hi / hi_roman / en), 0 missing |

## Blocking

### 1. All five datasets are already on ModelScope — the HF-gating note is incorrect

ModelScope is EvalScope's default `dataset_hub` ([`config.py`](https://github.com/modelscope/evalscope/blob/main/evalscope/config.py) → `dataset_hub: str = HubType.MODELSCOPE`), and every one of these datasets is already mirrored there under the **identical** ID, publicly, with the same config names and splits:

| dataset_id | ModelScope | Configs / splits verified |
| --- | --- | --- |
| `ai4bharat/MILU` | public | 11 language configs, `validation` 812 + `test` 14831 (Hindi) |
| `sarvamai/arc-challenge-indic` | public | 11 configs, `test` 1150 + `validation` 294 (hi) |
| `sarvamai/trivia-qa-indic-mcq` | public | 11 configs, `validation` 17944 (hi) |
| `sarvamai/boolq-indic` | public | single default config, `train` 103697 + `validation` 35970 |
| `sarvamai/gsm8k-indic` | public | 21 configs, `test` 1319 (hi) |

So MILU needs **no** token on the default path. Please drop these lines from `milu_adapter.py`:

```
- Requires access to the gated `ai4bharat/MILU` dataset — accept the dataset terms on
  huggingface.co and set `HF_TOKEN` (or run `huggingface-cli login`) before evaluating.
```

and replace them with a note that the dataset loads from ModelScope by default, and that `HF_TOKEN` + accepting the terms is only required when the user explicitly sets `dataset_hub: huggingface`. Same for the "Each language is a separate HF dataset config" phrasing in the other descriptions — please make it hub-neutral (e.g. "one dataset config per language").

### 2. Missing smoke tests

`AGENTS.md` requires a minimal runnable test for every new benchmark. Please add the five entries following the existing pattern in `tests/benchmark/test_llm.py`, so `record_to_sample` mapping and the answer-extraction path are covered.

### 3. Docs metadata is incomplete — we'll regenerate it on our side

The `_meta/*.json` files are half-populated for all five:

```
readme.zh = "" (needs_translation: true), statistics = {}, sample_example = {}
```

and consequently `docs/zh/benchmarks/*.md` are English copies. This is expected — `make docs-pipeline` needs `DASHSCOPE_API_KEY` for translation plus a dataset download for statistics, which you don't have. **No action needed from you:** once the description text in item 1 is final, we'll run

```
make docs-pipeline BENCHMARK="milu arc_indic triviaqa_indic indic_boolq gsm8k_indic" FORCE=1
make docs-generate
```

and push the regenerated `_meta` + `docs/{zh,en}` files. Please don't hand-edit those generated files in the meantime.

## Non-blocking

- **`eval.schema.ts` `optional()` → `nullish()` is a correct fix.** The backend returns `None` for an absent locale ([`service/utils/benchmarks.py`](https://github.com/modelscope/evalscope/blob/main/evalscope/service/utils/benchmarks.py) `parse_benchmark_description` guard), which serializes to JSON `null`, and `.optional()` only accepts `undefined` — it matches the existing `paper_url: z.string().nullish()` in the same file. Thanks for adding the test coverage too. That said, it's an independent contract fix unrelated to Indic benchmarks; ideally it would have been its own small PR (see the branching note below). Keeping it here is fine.
- **`DEFAULT_TEXT_BENCHMARKS`**: that list is a curated set of ~22 mainstream benchmarks surfaced in the Web UI (`gsm8k`, `mmlu`, `ceval`, `aime*`, …). Adding five language-specific benchmarks to the *default* list is a product decision — I'd prefer to leave the curated list unchanged for now and let users select these explicitly. Could you revert the `evalscope/service/utils/benchmarks.py` change?
- **Default evaluation volume is very large**: `triviaqa_indic` is ~17.9k × 11 ≈ 197k samples, `indic_boolq` 35970, `milu` well over 100k across 11 languages. Please mention in `## Evaluation Notes` that users should scope runs with `limit` and/or `subset_list`.
- **Dead code / small cleanups**:
  - `arc_indic_adapter.py:74` — the `answer_key.isdigit()` branch never triggers (all `answerKey` values are A–D across every config). Safe to keep as defensive, but it's untested by construction.
  - `milu`, `arc_indic`, `triviaqa_indic`, `gsm8k_indic` all define a no-op `__init__` that only calls `super().__init__(**kwargs)` — those can be removed (only `indic_boolq` needs one, for `reformat_subset`).
  - `LANGUAGE_NAMES` is duplicated verbatim in four adapters; a single shared constant would be cleaner.
  - `indic_boolq_adapter.py` hardcodes the English label `Question:` in front of a translated question, and appends `?` unconditionally. Intentional? For a multilingual benchmark it's worth a comment explaining the choice.
- **Few-shot**: `milu` sets `train_split=None` although a `validation` split (812) exists, and `indic_boolq` sets `train_split=None` although `train` (103697) exists. Fine for the 0-shot default, but wiring them up would let users run few-shot later.

## On maintaining a single long-lived branch

I'd rather stick with one branch per feature, and I don't think it will cost you more work — here's the reasoning:

1. **We squash-merge.** After #1569 lands as a single commit on `main`, a long-lived branch diverges structurally from `main`, and your next PR from that same branch will replay already-merged changes as new diff. That produces *more* conflicts than short-lived branches, not fewer.
2. **This PR already shows the coupling cost.** It contains five Indic adapters *plus* an unrelated Web schema fix. The schema fix is correct and could have been merged in a day; instead it's blocked behind the description/test/docs items above. One-topic branches mean the ready parts ship immediately.
3. **Review quality.** The branch currently carries two `Merge branch 'modelscope:main'` commits, which makes the diff noisier. If you need to sync with `main`, please `rebase` rather than `merge`.
4. **Your current workflow already works.** #1545 and #1563 were single-topic branches and both merged smoothly; #1570 is separate too. That's exactly the pattern I'd like to keep.

Practical suggestions to cut your overhead without a shared branch: branch every topic straight off `main`, use `git worktree add` so you don't pay a checkout cost when switching, and keep a local-only integration branch if you want to test several features together — just don't use it as a PR head. GitHub happily handles several concurrent PRs from one fork, so nothing serializes.

Happy to keep reviewing — please ping me once items 1 and 2 are addressed (and item 3 is on us).

Dhru1001:
Thanks for the thorough review, @Yunnglin — really appreciate the depth here (the per-adapter field-mapping checks especially).

Pushed a commit addressing both blocking items:

1. **HF-gating claim was wrong.** You're right — ModelScope is the default hub and all 5 datasets are already public there under identical IDs. Reworded milu's description: the HF_TOKEN/gating note now only applies if `dataset_hub` is explicitly set to `huggingface`. Also swapped "HF dataset config" for hub-neutral wording in milu/arc_indic/indic_boolq.
2. **Added smoke tests.** 5 `test_*_mock` methods in `tests/benchmark/test_eval.py`, following the existing `_run_dataset_test(..., use_mock=True)` pattern. Ran all 5 locally against real ModelScope data — passing.

Also picked up the non-blocking items:
- Reverted the `DEFAULT_TEXT_BENCHMARKS` addition — agreed that's a product call, not mine to make here.
- Added a note on milu/triviaqa_indic/indic_boolq that their default runs are large (up to ~198k samples) and can be scoped with `subset_list`/`limit`.
- Dropped the no-op `__init__` from milu/arc_indic/triviaqa_indic/gsm8k_indic (kept indic_boolq's, since it sets `reformat_subset`).
- Commented indic_boolq's hardcoded "Question:"/"?" scaffold so the English-vs-translated split reads as intentional.

One I left alone: the duplicated `LANGUAGE_NAMES` dict across 4 adapters. I don't see an existing pattern in the codebase for constants shared across sibling benchmark folders (every adapter's imports stay within its own folder), so I didn't want to invent one unprompted — happy to add it if you've got a preferred location in mind.

`make lint` and the full test suite are clean. Let me know if anything else needs another pass 


For branching I will take in the suggestions and implement it from my side

Yunnglin:
Docs regenerated for the five Indic benchmarks.

- Ran `make docs-pipeline BENCHMARK="milu arc_indic triviaqa_indic indic_boolq gsm8k_indic" FORCE=1` in the `eval` conda env.
- Updated `evalscope/benchmarks/_meta/*.json` with statistics + sample examples.
- Regenerated `docs/en/benchmarks/*.md` and `docs/zh/benchmarks/*.md` (including Chinese translations).

The generated docs are pushed to branch `docs-only/indic-language-benchmarks-1569` on `modelscope/evalscope`: https://github.com/modelscope/evalscope/tree/docs-only/indic-language-benchmarks-1569

I tried to push directly to this PR branch, but write access to `Bharatgen-Tech/evalscope` is denied. You can either cherry-pick/commit the docs files from the branch above, or we can merge that branch into main after this PR lands.

Code review notes (all addressed in the current PR head):
- HF gating claim fixed.
- Mock smoke tests added.
- `DEFAULT_TEXT_BENCHMARKS` reverted.
- Large-run warnings added.
- No-op `__init__` methods removed.
- indic_boolq scaffold comment added.
- Few-shot splits wired up for milu and indic_boolq.

All mock tests, `make lint`, `test_ci_lite`, and the web test suite pass locally.

Dhru1001:
@Yunnglin Done

Yunnglin (approved):
LGTM

All review comments addressed, docs regenerated, 7 mock tests + CI smoke test pass. Good to go.

PR #1570, opened 2026-08-12
Feature/report merge

Dhru1001:
## Summary
- Adds `docs/en/get_started/report_merge.md`, documenting the Merge and Rename report actions: how to trigger each from the report list UI, the validation rules Merge pre-checks (same model, disjoint datasets, no report tied to a running task), what a merge does under the hood (new run directory, copied artifacts, merged task config, regenerated HTML, source cleanup), and the full error/status code reference for both actions.
- Includes the raw `POST /api/v1/reports/merge` / `/rename` request/response shapes for scripting against the Web service directly.
- Registers the new page in the docs toctree (`docs/en/index.md`) and cross-links it from the Evaluation Report List section of `visualization.md`.


Dhru1001:
@Yunnglin

Yunnglin:
Hi @Dhru1001, thanks for the careful work here — path-safety with `ReportRef`, rollback on failure, tests, and bilingual UI copy are all appreciated, and CI is green.

I’m closing this because the feature mutates/deletes source evaluation runs, which conflicts with EvalScope treating reports as immutable audit records. The merged `task_config.yaml` only keeps the first source run’s config, so the report claims one setup produced scores that came from different settings; logs and `collection_detailed_report.json` are also dropped. The underlying need — seeing one model across multiple datasets — is better solved in the viewer layer, and the framework already supports it: resuming an existing run with an extended dataset list via `use_cache` writes everything into one directory with honest provenance.

Could you open an issue describing the original scenario instead? If the gap is that Compare requires a common dataset or the report list lacks model-level grouping, we can design a view-layer fix without data-loss risk. We can reuse the UI pieces from this PR once the behavior is agreed on.

Issue #1575, opened 2026-08-13
[Bug] LLM-judge failures are invisible in reports, and the generic score extractor can silently inflate scores (violating the documented fail-close contract)


YuhaoLin2005:
## Summary

The repo already documents a **fail-close** contract for LLM-judge failures: `hipho/utils.py:140-142` — *"A failed judge request must score 0, never full credit: `LLMJudge.judge` reports failures as an `[ERROR] ...` string."* Fail-close to 0.0 is therefore **intended behavior**, not the bug.

The actual gaps are fourfold, and they sit in the **shared** extraction path (`ScoreExtractor` → `LLMJudge.get_score`), which all judge-based benchmarks that don't override `llm_match_score` go through:

1. **The generic path can silently give *full credit* or *garbage credit* — the exact thing the fail-close contract forbids.**
   - `NumericScoreExtractor` clamps out-of-range values to `[0,1]` with **no warning** (`score_extractors.py:61`). A user running a 0–100-scale judge (custom `prompt_template`/`score_pattern`) gets `[[90]]` → **1.0**, i.e. the failed/extreme case is awarded *perfect* credit — the opposite direction of the documented "never full credit" intent.
   - Unlike `hipho/utils.py:144` (which checks the `[ERROR]` prefix **before** running any regex), `LLMJudge.get_score` (`llm_judge.py:178-190`) passes the raw response straight to the extractor. With a loose custom numeric pattern, digits embedded in the `[ERROR]` string (model_id `Qwen3-235B`, api_url `.../v1/`, etc.) can be parsed as a **garbage nonzero score**.
2. **A matched-but-unmapped category silently returns 0.0 with no warning** (`score_extractors.py:31`) — inconsistent with the no-match branch one line below (`:33`) which *does* `logger.warning`. Same file, same failure class, different observability.
3. **Judge failures are invisible in the aggregate report.** `score.explanation` preserves the raw `[ERROR]` text (`llm_judge_mixin.py:156`), but nothing downstream reads it — the default `Mean` aggregator folds every per-sample `{'acc': float}` into the headline number with no filtering, no failure counter, no report banner (`aggregators.py:34-52`). A judge outage silently drags the reported accuracy down.
[...]

PR #1576, opened 2026-08-13
fix(judge): fail closed on [ERROR] judge responses and surface silent extraction paths

YuhaoLin2005:
## Summary

Ports the **fail-close `[ERROR]` guard** that `evalscope/benchmarks/hipho/utils.py` already ships into the **shared** LLM-judge scoring path (`LLMJudge.get_score` → `ScoreExtractor`). When a judge API request fails, `LLMJudge.judge()` returns an `[ERROR] ...` string that embeds the model id, endpoint and HTTP code — under a loose numeric `score_pattern` (e.g. `(\d+)`) the shared extractor can parse those digits and clamp them to **full credit**, the exact thing the documented contract forbids ("A failed judge request must score 0, never full credit").

Also surfaces two **silent** paths in `score_extractors.py` with `logger.warning`, matching the observability the no-match branches already have:
- `PatternScoreExtractor`: a matched-but-unmapped answer returned `0.0` with no warning (inconsistent with the no-match branch one line below).
- `NumericScoreExtractor`: an out-of-range value (e.g. `[[90]]` from a 0–100-scale judge) was clamped to `[0, 1]` with no warning.

Adds unit tests locking in the documented fail-close contract (failed judge → `0.0`, never full credit). See issue #1575 for the audit that motivated this; the semantic question of *excluding* failed samples with a visible count stays open there.

## Changes

1. `evalscope/metrics/judge/llm_judge.py` — module constant `JUDGE_ERROR_PREFIX = '[ERROR]'`; `judge()` builds the failure sentinel from it; `get_score()` fails closed (logs + returns `0.0`) when the response starts with the prefix, before any regex extraction.
2. `evalscope/metrics/judge/score_extractors.py` — warning on matched-but-unmapped pattern answers; warning when a numeric value is clamped (clamp behavior itself unchanged).
3. `tests/metrics/judge/test_score_extractors.py` — 12 tests covering the fail-close contract, clamp behavior, the `[ERROR]` guard (with a loose pattern that would otherwise yield full credit) and the producer contract (`judge()` returns a prefix-prefixed string when `generate` raises).

## Behavior compatibility

For the **default** patterns (`^\s*([AB])\s*$` and `\[\[(\d+(?:\.\d+)?)\]\]`), `[ERROR]` responses already failed closed via the no-match path, so `get_score` output is unchanged — the guard only changes behavior for loose/custom patterns, which is the intended fix. Verified by execution.

## Tests

```
tests/metrics/judge/test_score_extractors.py ................ 12 passed
tests/api/test_default_data_adapter.py ......................... 5 passed
tests/benchmark/test_researchrubrics.py .......... 11 passed, 2 failed*
tests/benchmark/test_wide_search.py + tests/api/test_gdpval_benchmark.py + tests/perf/test_arguments_secrets.py ... 28 passed, 9 failed*
```

\* The 11 failures are **pre-existing environment issues** (missing optional deps such as `dateparser`, docker-related mocks) — confirmed identical on a clean `main` checkout (`git stash` + rerun → same 9/2 failures), unrelated to this change.

### DIFF
--- a/evalscope/metrics/judge/llm_judge.py
+++ b/evalscope/metrics/judge/llm_judge.py
@@ -9,6 +9,11 @@ from .score_extractors import NumericScoreExtractor, PatternScoreExtractor, Scor
 
 logger = get_logger()
 
+# Sentinel that ``judge`` returns instead of raising on a failed request.
+# Consumers must fail closed on it (score 0, never full credit) — the same
+# contract as ``evalscope/benchmarks/hipho/utils.py#JUDGE_ERROR_PREFIX``.
+JUDGE_ERROR_PREFIX = '[ERROR]'
+
 DEFAULT_PROMPT_TEMPLATE = """Your job is to look at a question, a gold target, and a predicted answer, and return a letter "A" or "B" to indicate whether the predicted answer is correct or incorrect.
 
 [Question]
@@ -159,7 +164,7 @@ class LLMJudge(BaseJudge):
         except Exception as e:
             error_message = f'Error occurred during {self.model_id}@{self.api_url} LLM judge evaluation: {e}'
             logger.error(error_message)
-            return f'[ERROR] {error_message}'
+            return f'{JUDGE_ERROR_PREFIX} {error_message}'
 
     def build_prompt(self, pred: str, gold: str, question: Optional[str] = None) -> str:
         if question is None:
@@ -186,5 +191,12 @@ class LLMJudge(BaseJudge):
             float: The numeric score extracted from the response
         """
         if response is None:
+            return 0.0
+        # A failed judge request must score 0, never full credit: ``judge`` reports
+        # failures as an ``[ERROR] ...`` string whose embedded digits (model id,
+        # endpoint, HTTP code) could otherwise be parsed as a score by a loose
+        # pattern. Same guard as ``benchmarks/hipho/utils.py``.
+        if response.startswith(JUDGE_ERROR_PREFIX):
+            logger.warning(f'LLM judge failed; returning 0.0. Response: {response}')
             return 0.0
         return self._score_extractor.extract(response)
--- a/evalscope/metrics/judge/score_extractors.py
+++ b/evalscope/metrics/judge/score_extractors.py
@@ -1,3 +1,4 @@
+import math
 import re
 from abc import ABC, abstractmethod
 from typing import Dict, Optional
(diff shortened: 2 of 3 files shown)


Yunnglin (approved):
LGTM.

Verified the root cause on `main`: `LLMJudge.get_score` passed the raw response straight to the extractor, so a failed judge request — which `judge()` reports as an `[ERROR] ...` string embedding the model id, endpoint and HTTP code — could be parsed as a score by a loose custom `score_pattern`. With `score_type='numeric'` and `score_pattern=r'(\d+)'` the embedded digits clamp to **1.0**, i.e. a failed request awarded full credit, which is the exact inverse of the contract `benchmarks/hipho/utils.py` already documents. `get_score` is the single call site for judge scoring, so the guard there covers the whole shared path.

Also confirmed the observability fixes are real gaps, not cosmetic: the matched-but-unmapped branch returned 0.0 with no log while the no-match branch one line below did warn, and out-of-range clamping was entirely silent.

Disclosure: I pushed one commit to this branch rather than round-tripping it, since CI was red and the issue was adjacent to the code being changed. `_clamped` treated NaN as in-range because every comparison against the bounds is false for NaN. `float()` accepts `'nan'`, so a loose pattern makes it reachable; on `main` that silently became 1.0, and after the refactor it became `nan`. Nothing in the metric aggregation path filters NaN, so one such sample would turn the whole dataset metric into NaN. It now fails closed to `clamp_min`. That commit also adds the blank lines yapf wanted, which was the failing pre-commit hook. Please review it as you would any other change here.

Non-blocking follow-ups, recorded so they are not mistaken for silent rejections:
- Copilot's suggestion to assert the new warnings is reasonable, but plain `caplog` captures nothing here: `utils/logger.py` sets `propagate = False`, so the records never reach the root handler and a naive assertion would pass without checking anything. The working idiom is the `caplog_evalscope` fixture in `tests/agent/external/test_responses_extended_item_types.py`.
- `tests/metrics/` is not covered by any CI job — `unit-tests` only runs `tests/cli/test_all.py::TestRun::test_ci_lite` — so these tests currently document the contract rather than enforce it. Worth wiring up separately.
- `JUDGE_ERROR_PREFIX` now exists in both `hipho/utils.py` and `metrics/judge/llm_judge.py`. Left as is to keep this diff minimal; it should collapse to one definition during the judge refactor.

Point 3 of #1575 — judge failures being invisible in the aggregate report — is deliberately out of scope here and will be handled as part of the broader judge-layer refactor described in that issue.

Thanks for the unusually well-evidenced issue and a fix that landed in the right layer.

PR #1584, opened 2026-08-15
feat(io): support `parquet` and binary image features in `general_vmcq`

Moenupa:
## Features

- Add a `guess_mimetype` toggle to `bytes_to_base64()` and `VisionLanguageAdapter._image_bytes_to_base64()` for auto-guessing the mimetype. Defaults to `False` for compatiblity.
  - Defaults to `False` so downstream functions are unimpaired.
  - If `True`, guess file types from the `bytes` object instead of given arguments. 
  - This is to facilitate general compatibility, as path-inferred types are unreliable.
- Add **parquet support** to the local dataset loader. Parquet reader preserves undecoded Hugging Face image/audio/video features (e.g. for image, read as `{"path": str | None, "bytes": bytes | None}`).

## Breaking Changes

- Modify General VMCQ adapter to accept columns with the following fallback order:
  1. if any `image_n` column is non-empty, use these columns;
  2. if `images` column is non-empty, use this as a list of images;
  3. otherwise return images as empty.
- Extend General VMCQ adapter to accept binary image payloads (in addition to `str`), regardless of column names:
  1. accept `str`, `dict`, and `None`, otherwise, raise `ValueError`;
  2. `dict` must be an undecoded [Hugging Face Image feature](https://huggingface.co/docs/datasets/en/about_dataset_features#image-feature), i.e., `{"path": str | None, "bytes": bytes | None}`, and auto-convert to image_url if binary payloads are provided.

## Tests

- Add [evalscope/api/benchmark/adapters/vision_language_adapter.py](evalscope/api/benchmark/adapters/vision_language_adapter.py) to test the above changes.

## Dependency

- Add `filetype` for guessing `bytes` to their mimetypes.


Moenupa:
Oddly enough, my local `make lint` fails to detect the double-quote issue, while GH action does.

Here's my `make lint` at https://github.com/modelscope/evalscope/pull/1584/commits/0d6a56cf939a4e22e0aba98f0ae5232a5b6f890b

```
pre-commit run --all-files
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

evalscope/perf/plugin/api/custom_api.py:194:5: F811 redefinition of unused 'Arguments' from line 5
evalscope/perf/plugin/datasets/custom.py:149:5: F811 redefinition of unused 'Arguments' from line 4

isort....................................................................Passed
yapf.....................................................................Passed
trim trailing whitespace.................................................Passed
check yaml...............................................................Passed
fix end of files.........................................................Passed
fix requirements.txt.....................................................Passed
fix double quoted strings................................................Passed
check for merge conflicts................................................Passed
mixed line ending........................................................Passed
```

I don't think `make lint` is robust enough for reproduction at this time. Do maintainers have any plan for improving it (migrating to ruff maybe)?

Yunnglin (changes requested):
Thanks for the contribution — the design is sound and the image path reuses the right helpers (`bytes_to_base64`, `compress_image_to_limit`, and `datasets`' `cast_column(..., decode=False)`). The `guess_mimetype` toggle is opt-in with a safe fallback and is backward compatible. However, there are two items that block merge plus a few cleanups.

## Blocking

- **CI lint failure — double-quote violations.** The new `_extract_images` / `_extract_videos` code uses double-quoted strings that `double-quote-string-fixer` rejects, so `run-pre-commit` fails on `general_vmcq_adapter.py`. Re your comment: this hook auto-fixes — its first run rewrites the file and exits non-zero, the second passes. Your local log shows `Passed`, meaning the hook already rewrote the file in your working tree but those fixes weren't committed, so CI still sees the double-quoted version. Just run `make lint`, then commit the modified files (`make lint` is identical to CI's `pre-commit run --all-files`).

- **Video payloads from Parquet are silently corrupted.** `parquet_to_list` explicitly undecodes `Video` features to `{"path", "bytes"}` dicts, but `_extract_videos` wraps the raw value as `{'url': <dict>, ...}`. Downstream, `_content_video_from_value` only recognizes `url`/`video`/`data` keys — an HF Video dict has none of them, so `video = video_value.get('url')` becomes the inner dict and `ContentVideo(video=str(dict))` produces a garbage path. Images correctly handle the `{"path","bytes"}` shape; videos do not, despite the PR advertising video support. Please either mirror the image handling for videos (with a test) or drop the video undecode from `parquet_to_list` to avoid a misleading half-feature.

## Suggestions

- **`BenchmarkMeta.description` not updated (docs drift).** The class docstring now mentions parquet/bytes dicts, but the `description` still says "Images/videos are plain strings (do not wrap in `{"url": ...}`)". Since `description` drives the auto-generated docs, please update it and run `make docs-pipeline BENCHMARK="general_vmcq" FORCE=1 && make docs-generate` (or leave the description intentionally minimal).

- **Third-party `filetype` import placed in the stdlib block** (`io_utils.py`, between `csv` and `hashlib`). isort passed only because the package isn't installed in the pre-commit env; once installed locally it will be moved to the THIRDPARTY group. Please move it now to avoid churn.

- **`images` list branch has no cap.** The `image_n` path is bounded by `MAX_IMAGES`, but `enumerate(record['images'])` is unbounded. A `MAX_IMAGES` guard would keep the two paths symmetric.

## Verdict

Requesting changes: the CI lint failure is a hard gate on `main`, and the video-from-parquet path is functionally broken while the PR claims to support it. Once lint passes and the video gap is fixed or explicitly scoped out, this is good to merge.

Moenupa:
- [x] CI lint issue was fixed in https://github.com/modelscope/evalscope/pull/1584/commits/2bc39e8303723111e845c03f9b10cf3b755eb465.
- [x] Drop support for `Audio` and `Video` features from parquets. I intend to solve this in another PR, which will address #1172.
- [ ] `BenchmarkMeta.description` was **NOT** updated.
  I do not have an API Key from DASHSCOPE (I guess this is for auto-translation) to perform this. I will leave this to the maintainers to fix.
- [ ] I intentionally did **NOT** fix `isort`. Frankly, the `isort` config simply does not make sense, and I don't intend to solve this by adding a yet another unsafe declaration on top of it. I will leave this to maintainers to figure this out.

## Other Fixes

> `images` column does not respect the cap defined by `MAX_IMAGES`. 

I add this feature **intentionally**, for a good reason: In some long-context benchmarks (OCR-based MCQ), images may exceed 100.

## A side problem: `isort` configs

I dont think this config is properly maintained.

https://github.com/modelscope/evalscope/blob/616e2c61b61a6f92b5fd1ceb9eea0a5e28b86719/setup.cfg#L8-L12

And for `io_utils.py`, the groups should be:

| builtin or 3rd‑party | deps (comma‑separated) |
|----------------------|------------------------|
| builtin | base64, csv, hashlib, io, json, datetime, os, re, string, unicodedata, typing |
| 3rd‑party | filetype, numpy, jsonlines, yaml, PIL (Image) |

```cfg
known_third_party = json,yaml # <- why is json here? json is built-in
# correct one is a big list of imports to be examined and marked as 3rd party
# known_third_party = filetype,jsonlines,numpy,... 
```

Overall, `isort` requires effort to set up and may conflict with other formatters. As for this repo, setup is not that great, with a lot of linters messing around. I really hope maintainers can gradually migrate CI to `ruff` instead.

Yunnglin (approved):
All blocking issues resolved — lint fixed, video/audio from parquet now raises `NotImplementedError` (no silent corruption), tests pass. Docs regenerated and pushed. LGTM, good to merge.

PR #1587, opened 2026-08-15
feat(metrics): Wilson CI + paired McNemar helpers

KeilerHirsch:
## Summary

Adds stdlib-only statistical helpers for fixed-item leaderboard reporting to `evalscope.metrics`:

- `wilson_ci(successes, n)` — 95% Wilson score interval for pass rates; well behaved near 0/1, unlike the normal approximation
- `paired_mcnemar(a_only, b_only)` — exact binomial two-sided p-value for comparing two models on the same fixed item set

Both are exported from `evalscope.metrics` and covered by unit tests in `tests/metrics/test_uncertainty.py`.

## Motivation

On fixed item sets (e.g. SWE-bench-style leaderboards), strict rank order overstates differences: the honest question for a #1-vs-#k gap is the paired McNemar test, and every aggregate should carry an uncertainty measure. These helpers make that reporting possible without new dependencies.

## Testing

- Helper values verified locally against reference values (Wilson 50/100 → (0.404, 0.596); McNemar (1,9) → p=0.0215); flake8 (project config) and isort are clean
- `pytest tests/metrics/test_uncertainty.py` — the repo's eager package init pulls the full framework dependency stack, so the local environment could only verify the helper values in isolation; the full test run is executed by upstream CI
- CI status: checks are pending on the head commit — no upstream Actions run has executed yet (fork-PR/first-contributor approval). This PR does not claim a green CI run.

### DIFF
--- a/evalscope/metrics/__init__.py
+++ b/evalscope/metrics/__init__.py
@@ -8,7 +8,7 @@ if TYPE_CHECKING:
     from .judge.llm_judge import DEFAULT_NUMERIC_SCORE_TEMPLATE, DEFAULT_PROMPT_TEMPLATE, LLMJudge
     from .math.parser import extract_answer, math_equal, strip_answer_string
     from .nlp.metrics import ExactMatch
-    from .utils.functions import bleu_ngram_one_sample, exact_match, macro_mean, mean, micro_mean, simple_f1_score
+    from .utils.functions import bleu_ngram_one_sample, exact_match, macro_mean, mean, micro_mean, paired_mcnemar, simple_f1_score, wilson_ci
     from .utils.rouge import compute_rouge_score, compute_rouge_score_one_sample, compute_rouge_score_one_sample_zh
     from .utils.text_normalizer import BasicTextNormalizer, ChineseTextNormalizer, EnglishTextNormalizer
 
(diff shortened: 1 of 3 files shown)


Yunnglin (commented):
Thanks for the clean, dependency-free implementation.

I verified the math locally and found no correctness issues: the Wilson interval matches the textbook values (50/100 → (0.4038, 0.5962), endpoints stay inside [0, 1] at 0/n and n/n), and `paired_mcnemar(1, 9)` = 0.021484375 is exactly `scipy.stats.binomtest(1, 10, 0.5)` — the doubled tail is exact here since the null binomial (p=0.5) is symmetric. All six tests pass.

The blocker is scope, not correctness: nothing calls these helpers. Grepping for `wilson_ci` / `paired_mcnemar` returns zero hits outside this PR, and the report layer has no notion of an interval or standard error today — so merging adds unreachable code that is simultaneously part of the public `evalscope.metrics` surface we then have to keep. That runs against the project's "no speculative code" rule.

Either way forward works for me:

1. Land them together with a real consumer — e.g. report-level uncertainty for pass-rate metrics, going through the existing `Metric` / report schema. That also pins down the intended semantics (which `n`, per-subset or aggregate).
2. Drop the top-level `evalscope.metrics` export for now and add it when the reporting feature that needs it lands.

Minor points, worth fixing regardless:

- `wilson_ci(...) -> tuple` should be `Tuple[float, float]`.
- `z` is undocumented, and the "95%" wording silently becomes wrong for a different `z`; other functions in this file use the `Args:` / `Returns:` form.
- No input validation: `successes > n` or negative counts silently return a meaningless interval, and `paired_mcnemar` with negative counts surfaces a raw `math.comb` ValueError.

PR #1589, opened 2026-08-17
fix: correct self.dataset typo in VQA.info

Ricardo-M-L:
Fix: fix: correct self.dataset typo in VQA.info


Yunnglin (approved):
LGTM. `self.datset` is the only occurrence of that misspelling in the class; every other reference (`__init__`, `createIndex`, `getQuesIds`, `getImgIds`) uses `self.dataset`, so `info()` would have raised `AttributeError` on any call. No code depends on the typo. Thanks for the fix!

PR #1591, opened 2026-08-17
fix(perf): guard None content in openai_api token accounting

tianba-sh:
## Problem
`''.join(choice_contents)` raised `TypeError` when a choice's text content was `None`.

## Trigger conditions (all three required)
1. API omits `usage` (self-hosted vLLM / some OpenAI-compatible endpoints)
2. A choice has null text content:
   - non-stream tool-call: `message.content = null` (result goes to `tool_calls`)
   - streaming tool-call: chunks carrying only `tool_calls`, `delta.content = null`
3. `--tokenizer-path` set → no usage → local token counting → content join

## Fix
Sanitize at the 5 fill sites (`openai_api.py:223,227,234,245,247`) with `.get(...) or ''`, mirroring the existing pattern in `openai_responses_api._collect_output_text`. The join at line 255 is left as-is to stay consistent with `openai_responses_api.py:70`.

## Tests
Added 2 unit tests in `tests/perf/test_openai_api_parse_responses.py` (no API key needed):
- `test_parse_responses_non_stream_tool_call_null_content`
- `test_parse_responses_stream_tool_call_null_delta_content`

All 10 tests pass; pre-commit (flake8/isort/yapf/mixed-line-ending) clean.

### DIFF
--- a/evalscope/perf/plugin/api/openai_api.py
+++ b/evalscope/perf/plugin/api/openai_api.py
@@ -220,18 +220,18 @@ class OpenaiPlugin(DefaultApiPlugin):
             return
         if response['object'] == 'chat.completion':
             for choice in response['choices']:
-                delta_contents[choice['index']] = [choice['message']['content']]
+                delta_contents[choice['index']] = [choice['message'].get('content') or '']
         elif response['object'] == 'text_completion':
             for choice in response['choices']:
                 if 'text' in choice and 'index' in choice:
-                    delta_contents[choice['index']].append(choice['text'])
+                    delta_contents[choice['index']].append(choice.get('text') or '')
         elif response['object'] == 'chat.completion.chunk':
             for choice in response['choices']:
                 if 'delta' in choice and 'index' in choice:
                     delta = choice['delta']
                     idx = choice['index']
                     if 'content' in delta:
-                        delta_contents[idx].append(delta['content'])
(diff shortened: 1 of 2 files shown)


Yunnglin (approved):
LGTM.

Verified the fix locally: checked out the PR branch, then reverted only `evalscope/perf/plugin/api/openai_api.py` to `main` while keeping the new tests — both new tests fail with `TypeError: sequence item 0: expected str instance, NoneType found` at `openai_api.py:255`. With the fix applied, all 10 tests pass. So these are genuine regression tests, not self-confirming assertions.

Blast radius confirms this is worth fixing: `parse_responses` is called from `BenchmarkData.finalize`, which runs inside `BenchmarkMetrics.update()`'s success branch with no surrounding `try` — so the `TypeError` aborted the whole benchmark aggregation, not just one request.

Two follow-ups, neither blocking this PR:

1. Tool-call responses without a `usage` block now count `output_tokens = 0`, since `tool_calls[].function.arguments` are real emitted tokens that never reach `delta_contents`. Strictly better than crashing, but it's silent under-reporting of throughput — worth a separate issue rather than expanding this fix.
2. `custom_api.py:148-150` has the same class of unguarded access. Out of scope here per the minimal-changes convention.

Nice choice to sanitize at the fill sites rather than at the `join` — keeps the `List[str]` contract intact instead of masking future type pollution downstream.

PR #1597, opened 2026-08-18
fix(multi-choice): parse the last valid answer label instead of the first

juzihan0459:
## Description

Fixes multi-choice answer extraction when a reasoning model emits several `ANSWER:` markers before the final answer.

**Problem:** `parse_answers()` used `re.search` and kept the **first** `ANSWER:` match. Reasoning models often restate the required format (`'ANSWER: [LETTER]'`) or hedge with sentences like `Final answer: ANSWER: D` before giving the real answer. The first marker can then capture a placeholder or a partial word (e.g. `ANSWER` from `answer: ANSWER: D`), causing the sample to be scored as a miss even though the model's final choice is present later in the output.

**Fix:** iterate over **all** `ANSWER:` markers and keep the **last** one whose capture is a valid label; the same rule is applied to the Chinese parser (`parse_answers_zh`). A wrong letter leaked into the chain of thought can no longer shadow the final answer.

## Verification

- New regression tests cover: placeholder restated in reasoning, `Final answer: ANSWER: D` hedges, leaked letters inside the chain of thought, and the Chinese counterpart.
- `tests/test_multi_choices.py`: 15 passed.
- On a real LongBench-v2 run, this restores 13/25 previously-empty samples (accuracy 57.0% → 64.2%).
- `flake8`, `isort`, `yapf` all clean.

## Related

Fixes answer extraction failures observed with verbose reasoning-model outputs (e.g. DeepSeek-V4-Flash).

### DIFF
--- a/evalscope/utils/multi_choices.py
+++ b/evalscope/utils/multi_choices.py
@@ -165,8 +165,7 @@ _BRACKETED_LABEL_RE = re.compile(r'[\(\[（【]\s*([A-Za-z\d](?:\s*[,，]\s*[A-Z
 _PLAIN_LABEL_RE = re.compile(r'([A-Za-z\d][A-Za-z\d ,]*)')
 _PLAIN_LABEL_ZH_RE = re.compile(r'([A-Za-z0-9][A-Za-z0-9,，]*)')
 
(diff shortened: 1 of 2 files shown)


Yunnglin (approved):
Looks good to me.

PR #1604, opened 2026-08-19
feat(web): expose output directory name in Eval/Perf task forms

Dhru1001:
Lets users name a run's output directory from the Tasks tab instead of relying on the auto-generated timestamp id. The name becomes the task id (sanitized to filesystem/header-safe characters), which is what the backend already uses to derive work_dir, progress tracking, log lookup, resume, and report listing -- so no backend changes are needed and none of those flows break.

### DIFF
--- a/evalscope/web/src/components/eval/EvalConfigForm.tsx
+++ b/evalscope/web/src/components/eval/EvalConfigForm.tsx
@@ -9,12 +9,12 @@ import {
   validateDatasetArgs,
   FORM_MESSAGE_KEYS,
 } from '@/domain/form/validation'
-import { ApiKeyField, ApiUrlField, ModelField } from '@/components/tasks/TaskFormFields'
+import { ApiKeyField, ApiUrlField, ModelField, OutputDirField } from '@/components/tasks/TaskFormFields'
 import TaskFormShell from '@/components/tasks/TaskFormShell'
 import { useTaskForm } from '@/components/tasks/useTaskForm'
 
 interface Props {
-  onSubmit: (config: Record<string, unknown>) => void
+  onSubmit: (config: Record<string, unknown>, outputDirName?: string) => void
   disabled?: boolean
   initialDataset?: string
   /**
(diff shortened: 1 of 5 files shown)


Dhru1001:
@Yunnglin

Yunnglin:
Thanks for adding this. The user need is reasonable, but the implementation should reuse EvalScope’s existing output configuration instead of turning the directory name into `task_id`.

EvalScope already has `TaskConfig.work_dir` / `no_timestamp` for Eval and `outputs_dir` / `name` / `no_timestamp` for Perf. Please pass the setting through the normal config contract and let the service resolve it safely under `EVALSCOPE_OUTPUT_DIR`.

`task_id` must remain independently unique because it identifies processes, progress, logs, reports, and stop operations. With the current stable `name -> task_id` mapping:

- running the same Perf name twice fails because `benchmark_data.db` already exists;
- concurrent same-name tasks overwrite the same process-registry entry and share one output directory.

Please separate the user-facing output/run name from the unique task ID, define collision/resume behavior explicitly, and add coverage for repeated names, concurrent submissions, Perf reruns, Unicode names, and path traversal. If this is only for readability, the simplest solution is a readable name plus a unique suffix.

Dhru1001:
You're right, and I traced it through the code to confirm both failure modes:

- `outputs_dir = OUTPUT_DIR/task_id` (`evalscope/service/blueprints/perf.py`), and `benchmark_data.db` lives inside that directory (`evalscope/perf/utils/db_util.py:138`). `task_id` here is just `prefix_<sanitized-name>` with no uniqueness suffix, so a second Perf run given the same name reuses the exact same `outputs_dir` and DB path.
- `register_process()` in `evalscope/service/utils/process.py` keys `_active_processes` by `task_id` with a plain dict assignment, so two concurrent submissions with the same name silently overwrite each other's registry entry — the first process becomes unstoppable and untracked.

That's a real correctness gap, not a style preference, so this needs the redesign you described rather than a patch on top. Closing this PR — I'll rework it to route the display name through `TaskConfig.work_dir`/`no_timestamp` (Eval) and `outputs_dir`/`name`/`no_timestamp` (Perf) instead of folding it into `task_id`, keep `task_id` independently unique, and add coverage for repeated names, concurrent submissions, Perf reruns, unicode names, and path traversal before reopening. Thanks for catching this.

Issue #1599, opened 2026-08-18
Report list: group same-model reports for display, without merging report directories


Dhru1001:
### Context

Our eval workflow runs one model against datasets incrementally—often as separate invocations (different dataset, different day) rather than one big run. The report list therefore ends up with N rows per model. For example, a model evaluated on 6 datasets across 3 separate runs appears as 3 unrelated rows, with no way to see the model's full picture without opening each report individually.

We tried solving this by physically merging report directories in #1570. @Yunnglin correctly closed it:

> I'm closing this because the feature mutates/deletes source evaluation runs, which conflicts with EvalScope treating reports as immutable audit records. The merged `task_config.yaml` only keeps the first source run's config, so the report claims one setup produced scores that came from different settings; logs and `collection_detailed_report.json` are also dropped. The underlying need—seeing one model across multiple datasets—is better solved in the viewer layer, and the framework already supports it: resuming an existing run with an extended dataset list via `use_cache` writes everything into one directory with honest provenance.
>
> Could you open an issue describing the original scenario instead?

This is that issue, along with a working alternative.

### Going forward vs. already-split history

For *new* runs, `use_cache` + an extended `datasets` list already does the right thing (confirmed).

However, this doesn't help with runs that are already split across directories, nor does it provide a fast way to see cross-dataset coverage per model in the list UI generally. Using `use_cache` in this case would mean re-evaluating every dataset that isn't already present in whichever single directory you resume from, since `use_cache`'s cache lookup is scoped to one `work_dir`.

### Proposed / implemented: view-layer grouping, no file mutation

We put together a display-only alternative and would like feedback before opening a PR:

* `GET /api/v1/reports?group_by=model` rolls same-model report-list items into one row each, entirely in memory, using metadata the endpoint already loads for the flat list (`_build_report_meta` results grouped by `model_name`). No report is read, written, moved, or merged.
* Each group carries `children`: every constituent report exactly as the flat list describes it, with its own `run_id`, `model_id`, and `primary_metrics`. This means a group never shows a fabricated combined score.
* If a dataset appears in more than one of the model's reports, there is no silent "winner"; both remain visible under their respective reports when the group is expanded.
* The report list UI adds a **Group by model** toggle that renders these as expandable rows, reusing the existing `ReportsTable` / `ReportCard` components for the expanded children, so no new row-rendering logic is needed.
[...]

PR #1607, opened 2026-08-20
perf(service): memoize report-list metadata and support conditional GET

Yunnglin:
## Motivation

The `GET /api/v1/reports` endpoint read and parsed **every** report's JSON (plus each run's YAML config) on **every** request, then sliced the result to a single page. That is `O(all reports)` of disk + parse work behind a paginated contract, repeated on every call with no caching — the frontend consuming one page never saved the backend that cost.

The paginated API contract is correct; only the work behind it was wasteful. This PR fixes the work, leaving the contract (and the frontend) unchanged.

## Changes

**Layer 0 — read only what the list needs**
- `_build_report_meta` now builds list metadata from the report JSONs alone, dropping the unused `load_report_bundle` (which also parsed the run's YAML).
- Fixes a real bug: a report directory with no `configs/*.yaml` was silently dropped from the list, because `load_report_bundle` raised `FileNotFoundError` and the error was swallowed.

**Layer A — memoize per-reference metadata (`report_meta_cache.py`)**
- Cache keyed by a stat-only file fingerprint `(relpath, mtime_ns, size)`. Reports are immutable audit records (a rerun rewrites files → mtime changes; a delete removes the dir), so an unchanged fingerprint guarantees unchanged metadata — invalidation is free.
- Keyed by `(outputs root, reference)` so switching `root_path` never evicts or shadows another root's warm cache; entries for a vanished reference are pruned.
- Each request now stats all references but reads/parses only the new or changed ones.

**Layer B — conditional GET**
- The list response carries an `ETag` (over the scanned fingerprints + query params) and `Cache-Control: no-cache`. Browsers revalidate and reuse an unchanged page as a `304`. No frontend change: the client only ever sees `200` + body (the browser serves the cached body transparently).

**Housekeeping**
- Split `list_reports` into focused helpers (`_load_report_metas`, `_apply_report_filters`, `_list_response`).
- Exception handlers now capture stack traces (`exc_info=True`) and return a generic message instead of echoing raw exception text to clients.
- Added return-type annotations to the view functions.

## Notes / scope

- Response JSON shape is **unchanged**; this is purely a backend improvement. No frontend files touched.
- A full persistent index (sidecar/SQLite) was intentionally **not** added — the remaining per-request cost is a cheap stat-scan, and an index would be premature for the current scale.
- This lands the backend groundwork for a future view-layer "group by model" feature (#1599), which will build on this cheap metadata path.

## Testing

- `tests/report/` — all green except one pre-existing failure on `main` (`olmocr_bench_adapter` `unit_test_pass_rate` vocabulary), unrelated to this change.
- Added endpoint tests: config-less report is listed (Layer 0), cache hit / fingerprint invalidation / prune (Layer A), and conditional `304` (Layer B).
- Verified end-to-end against a real outputs directory, both via the Flask test client and a live `app.run` server: `200` + `ETag` + `Cache-Control` on first request, `304` (empty body) on `If-None-Match`, multi-root alternation stays warm, filters/sort intact, bad params rejected with `400`.

### DIFF
--- a/evalscope/service/blueprints/reports.py
+++ b/evalscope/service/blueprints/reports.py
@@ -12,18 +12,26 @@ import plotly.graph_objects as go
 import shutil
 from datetime import datetime
 from flask import Blueprint, jsonify, request, send_file
+from flask.typing import ResponseReturnValue
 from pydantic import ValidationError
 from typing import List, Optional, Tuple
 
 from evalscope.constants import PLOTLY_CDN_URL, PLOTLY_THEME
 from evalscope.metrics.semantics import PrimaryMetricRef
-from evalscope.report import ReportKey, ReportRef, get_data_frame
+from evalscope.report import ReportKey, ReportRef, get_data_frame, get_report_list
 from evalscope.report.report import Report
 from evalscope.report.visualization import (
     plot_multi_report_radar,
     plot_single_dataset_scores,
     plot_single_report_scores,
     plot_single_report_sunburst,
+)
+from evalscope.service.report_meta_cache import (
+    Fingerprint,
+    build_report_meta_cached,
+    list_etag,
+    prune_report_meta_cache,
+    report_ref_fingerprint,
 )
 from evalscope.utils.data_utils import (
     get_acc_report_df,
@@ -36,6 +44,7 @@ from evalscope.utils.data_utils import (
     load_multi_report_groups,
     load_report_bundle,
     normalize_score,
+    report_model_dir,
     scan_report_refs,
 )
 from evalscope.utils.io_utils import OutputsStructure
@@ -78,7 +87,7 @@ _MEDIA_EXTENSIONS = {
 
 
 @bp_reports.route('/media/file', methods=['GET'])
-def serve_media_file():
+def serve_media_file() -> ResponseReturnValue:
     """Serve a local media file (image / audio / video) via HTTP.
 
     This proxy endpoint allows the browser to load server-side local file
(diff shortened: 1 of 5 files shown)


Yunnglin:
Thanks for the review — all three addressed in 571448fc:

1. **`list_etag` O(N) intermediate payload** — now hashes incrementally over the sorted `(ref, fingerprint)` pairs (and query parts), so no large joined string is allocated. Ordering is still deterministic (keys are unique per scan).
2. **`_build_report_meta` swallowing exceptions silently** — it now logs the ref key with `exc_info=True` before returning `None`. Since the `None` result is cached, this logs on a cache miss rather than on every request, so it aids diagnosis without spamming.
3. **`_CACHE` growth across roots** — clarified the comment: bounded per root (pruned every request) and, across roots, growing only with the number of distinct outputs roots the process serves. The list endpoint rejects non-existent roots, so in practice this set is small and stable; no cross-root eviction policy added to avoid premature complexity for the current scale.

Yunnglin:
Follow-up self-review fixes landed in 0b897b09:

- Fingerprints now include the run-directory mtime, because non-timestamp run IDs derive their displayed timestamp from that mtime. They also include file ctime to invalidate on permission/metadata-only changes.
- Failed metadata (`None`) is no longer cached, so transient read/parse failures can recover without requiring a fingerprint change.
- Cache publication now uses compare-and-set semantics: a slow computation based on an older fingerprint cannot overwrite a newer result published concurrently.
- Added focused regression tests for all three cases.

Validation: 23 focused tests pass; real-output smoke remains `SMOKE_OK` (200/ETag, 304 empty body, sort/validation, multi-root A/B/A). The full `tests/report/` result is 696 passed plus the same pre-existing `olmocr_bench_adapter` aggregation-vocabulary failure seen on the local `origin/main` baseline.

PR #1609, opened 2026-08-20
feat(io): support media placeholders in `general_vqa`

Moenupa:
## Summary

- [x] (feat) Add `VisionLanguageAdapter._resolve_media_placeholders()` for resolving media placeholders.
  - `[{"role": "...", "content": "..."}]` triggers resolution.
    Returns `[{"role": "...", "content": list[Content]}]`
  - `[{"role": "...", "content": [{"type": "text", "text": "..."}, ...]}]` bypasses this resolution.
    Returns itself without modification.
- [x] (feat) Add a bypass for `content_from_openai()` to skip already-parsed list of content. 
- [x] (break) `GeneralVQAAdapter.record_to_sample()` now performs a media placeholder resolution before parsing from OpenAI format. Resolution directly outputs each messages' content as `list[Content]` instead of OpenAI format, which is why the bypass at `content_from_openai()` is needed.
- [x] Update Documentation for `GeneralVQAAdapter` once we agree on the breaking change.
- [x] Run docs generation pipeline.

## Tests

```sh
# api changes
pytest tests/api/test_vision_language_adapter.py
# general_vqa
pytest tests/benchmark/test_general_vqa_adapter.py
# backward compat
tests/cli/test_custom.py
```


Yunnglin:
PR reuses existing media-parsing utilities well, but one of its own tests is failing:

- `TestResolveMediaPlaceholders::test_system_message_preserved` fails because `_resolve_media_placeholders()` transforms string content for **all** message roles, while the test expects only user messages to be modified.

This also feels like over-generalization: placeholders like `<image N>` only make sense in user messages, so resolving them for system/assistant roles is surprising and unnecessary.

Moenupa:
Yes. And this is worth a small discussion:

In agentic scenarios, `user` is not the only party that emits media artifacts.
Tools can also yield media.

We can enforce role checking for media placeholders to be injected:
```py
# Option A/inclusion, triggers only for 'user', this is bad for agents, '<image n>' may be inside tool
if msg.get('role') == 'user': ...

# Option B/exclusion, bypass for certain roles, also not good, needs documentation and user notice
# hardcoding would require endless expansion in future and not generally reliable
if msg.get('role') != 'system': ...
```

But I feel both are bad for compatibility, and I dont really have a good proposal so far. I would like to hear your advice on how we move forward with this feature.

I will hold this PR until a consensus is reached.

Moenupa:
Just read the `openai` docs and found only user chat msgs support multimedia. I will go ahead with implementation of option A and improve the tests.

Moenupa:
@Yunnglin Fix the implementation and improve tests.

Media placeholder feature now only works if both:
1. `msg['role'] == 'user'`
2. `isinstance(msg['content']) == 'str'`

If you agree, I will next update the docs accordingly.

Yunnglin:
Thanks! The fix looks correct — verified locally, all 32 tests pass and non-user messages (system/assistant/tool) are now left untouched. Agreed on the approach, please go ahead with the docs update.

Issue #1605, opened 2026-08-19
Follow-up to #1595: migrate VL adapters to _extract_media & harden parquet media passthrough


Yunnglin:
## Summary

PR #1595 introduces a unified media loader `VisionLanguageAdapter._extract_media()` / `_normalize_media_value()` that resolves image / video / audio cells into a common `{'url', 'format', **extra}` payload. Two follow-ups are intentionally left out of that PR to keep its diff minimal; this issue tracks them.

## 1. Migrate the remaining VL adapters to `_extract_media`

`mmmu`, `mmmu_pro`, `cmmmu` and `blink` still hand-roll their own `image_k` loops plus `bytes_to_base64` before calling `_parse_text_with_images`, which is exactly the duplication the new `_extract_media` was built to remove.

- `evalscope/benchmarks/mmmu/mmmu_adapter.py`
- `evalscope/benchmarks/mmmu_pro/mmmu_pro_adapter.py`
- `evalscope/benchmarks/cmmmu/cmmmu_adapter.py`
- `evalscope/benchmarks/blink/blink_adapter.py`

**Why not in #1595:** migrating them changes their prompt-building path (the unified loader adds `max_image_bytes` compression and MIME guessing), so it can shift rendered prompts and therefore benchmark scores. That needs a before/after score comparison on at least `mmmu` and must not ride along in an IO PR. Each migration should ship with a parity check (same `content_list` shape on a fixed sample) rather than a blind swap.

## 2. `parquet_to_list` no longer fails loudly on undecoded Audio/Video columns

Before #1595, `evalscope/utils/io_utils.py::parquet_to_list` raised `NotImplementedError` for HF `Audio` / `Video` columns. #1595 removes that guard (`cast_column(..., decode=False)`) so `general_vmcq` can consume the raw `{'bytes': ...}` / `{'path': ...}` dicts.

**Side-effect:** any adapter *other than* `general_vmcq` that reads a local parquet with an HF Audio/Video column now silently receives those undecoded dicts instead of a clear "not supported" error. If such an adapter does not expect the dict shape, the failure now surfaces far from the root cause. We should either (a) document that `parquet_to_list` returns undecoded media dicts and have consumers validate, or (b) gate the passthrough behind the adapter's declared media support.

## Notes

- No functional regression in #1595 itself — the empty-cell / wav-format / stale-path / format-validation bugs found during review are fixed and covered in `tests/api/test_vision_language_adapter.py`.
[...]

PR #1611, opened 2026-08-20
refactor(benchmark): use shared media loader for BLINK images

lcache666-svg:
### What

- Reuse `VisionLanguageAdapter` media normalization in the BLINK adapter.
- Accept undecoded image bytes, local paths, URLs, and plural `images` records.
- Add adapter tests covering byte and list-based image inputs.

### Why

BLINK previously assumed every image field was a Hugging Face byte dictionary and encoded it directly. That bypassed the shared media handling used by other vision-language benchmarks and made path, URL, and list-based records inconsistent.

### Validation

- `python -m compileall -q evalscope/benchmarks/blink/blink_adapter.py tests/benchmark/test_blink_adapter.py`
- `git diff --check`
- `pytest` was not available in the current environment; the test file is included but has not been executed here.

### DIFF
--- a/evalscope/benchmarks/blink/blink_adapter.py
+++ b/evalscope/benchmarks/blink/blink_adapter.py
@@ -1,16 +1,11 @@
-import re
 from typing import Any, Dict, List
 
 from evalscope.api.benchmark import BenchmarkMeta, MultiChoiceAdapter, VisionLanguageAdapter
 from evalscope.api.dataset import Sample
-from evalscope.api.messages import ChatMessageUser, Content, ContentImage, ContentText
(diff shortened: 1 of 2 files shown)


Moenupa:
I dont think this PR is ready to merge, for a few reasons:

1. The efforts seem too superficial. `_extract_media()` may return None values, which both `src` and `tests` overlooked.
2. The `tests` are fake and hard-coded, insufficient to indicate no shifts after code refactoring.
3. #1605 explicitly states "That (i.e., refactoring media loading) needs a before/after score comparison"; this PR fails to address even "data input stays the same before/after refactoring".

PR #1617, opened 2026-08-21
feat: support text-to-image API services

### LINKED ISSUE #1040: 文生图模型无法调用本地部署的服务
## 问题描述

我在本地部署了一个Qwen-Image服务：http://localhost:8895/v1/images/generations
现在想用evalscope来评测，无法访问该链接

## EvalScope 版本（必填）
v1.2.0

## 执行的代码或指令

nohup evalscope eval \
  --model /data02/Qwen-Image \
  --api-url http://localhost:8895/v1/images/generations \
  --model-task image_generation \
  --api-key EMPTY \
  --eval-type openai_api  \
  --datasets evalmuse genai_bench general_t2i hpdv2 tifa160 \
  --eval-batch-size 1 \
  --timeout 60 \
  --stream \
  --limit 10 \
  > evalscope-Qwen-Image.log 2>&1 &

## 错误日志

File "/data01/lht/lht_env/lib/python3.11/site-packages/openai/_base_client.py", line 1259, in post
    return cast(ResponseT, self.request(cast_to, opts, stream=stream, stream_cls=stream_cls))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data01/lht/lht_env/lib/python3.11/site-packages/openai/_base_client.py", line 1047, in request
    raise self._make_status_error_from_response(err.response) from None
openai.NotFoundError: Error code: 404 - {'code': 400001, 'message': '请求路径错误', 'details': '请求路径错误'}

发现它会在请求的路径上拼上/chat/completions

若将--eval-type openai_api改为service，则显示不存在service参数：ValueError: Model API 'service' is not registered. Available model APIs: ['mock_llm', 'openai_api', 'server', 'llm_ckpt', 'checkpoint', 'text2image', 'image_editing']

若将evalscope版本降至0.17.0，则会报错：

File "/data01/lht/lht_env/lib/python3.11/site-packages/evalscope/metrics/t2v_metrics/__init__.py", line 44, in fga_blip2_score
    from .itmscore import ITMScore
  File "/data01/lht/lht_env/lib/python3.11/site-packages/evalscope/metrics/t2v_metrics/itmscore.py", line 4, in <module>
    from .models.itmscore_models import get_itmscore_model, list_all_itmscore_models
  File "/data01/lht/lht_env/lib/python3.11/site-packages/evalscope/metrics/t2v_metrics/models/itmscore_models/__init__.py", line 2, in <module>
    from .blip2_itm_model import BLIP2_ITM_MODELS, BLIP2ITMScoreModel
  File "/data01/lht/lht_env/lib/python3.11/site-packages/evalscope/metrics/t2v_metrics/models/itmscore_models/blip2_itm_model.py", line 8, in <module>
    from ..vqascore_models.lavis.models import load_model
  File "/data01/lht/lht_env/lib/python3.11/site-packages/evalscope/metrics/t2v_metrics/models/vqascore_models/lavis/__init__.py", line 13, in <module>
    from .models import *
  File "/data01/lht/lht_env/lib/python3.11/site-packages/evalscope/metrics/t2v_metrics/models/vqascore_models/lavis/models/__init__.py", line 15, in <module>
    from .blip2_models.blip2 import Blip2Base
  File "/data01/lht/lht_env/lib/python3.11/site-packages/evalscope/metrics/t2v_metrics/models/vqascore_models/lavis/models/blip2_models/blip2.py", line 26, in <module>
    from .Qformer import BertConfig, BertLMHeadModel
  File "/data01/lht/lht_env/lib/python3.11/site-packages/evalscope/metrics/t2v_metrics/models/vqascore_models/lavis/models/blip2_models/Qformer.py", line 27, in <module>
    from transformers.modeling_utils import (PreTrainedModel, apply_chunking_to_forward, find_pruneable_heads_and_indices,
ImportError: cannot import name 'apply_chunking_to_forward' from 'transformers.modeling_utils' (/data01/lht/lht_env/lib/python3.11/site-packages/transformers/modeling_utils.py)

## 运行环境

- 操作系统：ctyunos 23.01 2
- Python版本：3.11.14

## 其他信息

希望可以提供一个评测本地部署的http://localhost:8895/v1/images/generations链接的文生图精度评测方式，谢谢！

Excelius-Wang:
## Summary

- support OpenAI-compatible image generation services for text-to-image evaluation
- automatically select the `text2image` backend for image generation tasks configured with `api_url`
- keep Diffusers dependency checks scoped to local pipeline usage
- parse base64 and URL image responses while preserving response metadata
- add regression coverage for request construction, response parsing, error handling, config dispatch, and local dependency behavior

## Testing

- `pytest tests/models/test_text2image_api.py -q` — 5 passed
- `pytest tests/models -q` — 51 passed
- `pytest tests/cli/test_all.py::TestRun::test_ci_lite -q -p no:warnings` — 1 passed
- targeted pre-commit hooks on all changed files

Closes #1040


Yunnglin (changes requested):
## Review: PR #1617 — Requesting changes

The direction is right and #1040 needs this, but the happy path is the only path that works: hosted providers fail on auth, transient errors discard samples, and the design doesn't extend to the adjacent cases it will immediately be asked to cover.

### 1. Design & extensibility

- **`openai` is already a hard dependency** (`requirements/framework.txt`, imported unguarded in `openai_compatible.py`). Hand-rolling the HTTP layer re-implements `client.images.generate()`, and the net effect is that we **lose retries, lose structured server error bodies, and introduce an auth convention that disagrees with `OpenAICompatibleAPI` 100 lines away**.
- **`Text2ImageAPI` is now a two-personality class**: the local and service branches set two entirely disjoint sets of attributes and share nothing but the class name. The `model_args` semantic collision below is the direct consequence of two unrelated state machines sharing one constructor.
- **The dispatch mechanism is a third pattern in this repo.** Registry + `eval_type` is the framework's first-class backend selector; `text2speech` at least dispatches explicitly on `model_args.provider`. Inferring the backend from "is `base_url` truthy" is a parallel mechanism.
- **What does the next request cost?** When `image_editing` gets the same treatment (`/v1/images/edits`), **nearly all of the ~90 lines of HTTP plumbing here has to be copy-pasted**, because every piece of it is a private method on this one class. A non-OpenAI response shape has no hook, and an async submit-and-poll protocol has nowhere to live. One case is solved with no reuse surface for the three or four adjacent ones.
- **Flexibility points the wrong way:** four channels can inject fields into the request; the response parser has zero extension points. We can send to services we then cannot parse.

Going through the SDK settles the extensibility and the simplicity concerns together: the service branch shrinks to ~30 lines, `images.edit` later reuses the same client, retry policy and error semantics — no new abstraction layer, the reuse surface is the SDK.

### 2. Correctness bugs

1. The default `api_key='EMPTY'` is sent verbatim as `Authorization: Bearer EMPTY`, with no env-var fallback — hosted providers 401.
2. **No retries**: `retries` / `retry_interval` are present in the config and ignored, so one transient 429 discards a sample.
3. Errors omit the server's response body — precisely the only useful diagnostic in the issue this PR closes.
4. Local-mode `model_args` (`pipeline_cls` and friends) leak into the HTTP request body.
5. `ModelOutput.time` contradicts itself within one method: the service branch is correct, the local branch still returns an absolute epoch.

### 3. Functional coverage gaps

1. **DashScope / Model Studio image generation cannot be evaluated** — it doesn't support compatible-mode — yet the tests use `'Qwen-Image'` as the model name. It's the first thing users will try.
2. Services keyed `images` instead of `data` (e.g. SiliconFlow) fail to parse, with no workaround available to the user.
3. Only the `/images/generations` suffix is normalised; the more commonly pasted `/chat/completions` root is not, though we already handle that for chat.
4. `num_inference_steps` / `guidance_scale` become silent no-ops in service mode — and the framework's **default** image-gen config ships both.
5. Multi-image responses are silently truncated to the first image; the URL branch does no content validation, so a bad download only surfaces later at metric time.
6. The sibling half of the same feature — `image_editing` over a service — is untouched.

### 4. Test gaps

Only the `data: []` branch is covered. Non-2xx responses, invalid base64, data-URI-prefixed `b64_json`, retry behaviour, and **the default `EMPTY` api-key header** are untested — the tests use `api_key='test-key'` / `None`, which is exactly what skips the real CLI default. "Regression coverage for … error handling" in the description overstates what's there.

### 5. Docs gaps

No doc changes: `parameters.md` still describes `text2image` as "(diffusers)", `t2i.md` has no service recipe (the issue explicitly asks for one), and **the compatibility boundary is undeclared** — including that the most obvious provider to try doesn't work.

### 6. Scope decision

Which bar this is held to should be settled first:

- **As a narrow fix for #1040 (self-hosted, OpenAI-shaped endpoints):** fix 2.1–2.4 and 3.3, state the supported shape in the docs, and it lands.
- **As a framework-level "T2I via API service" capability:** not yet — the design doesn't extend, the compatibility boundary is unstated, and the docs are absent, so the next contributor either copies the plumbing or refactors it away.

I'd suggest targeting the first explicitly and scoping it in the docs, rather than half-landing the second.

### Suggested order

1. Settle the SDK question — it subsumes 2.2 and 2.3.
2. Then the rest of §2, plus 3.1–3.3.
3. Then the missing tests and an explicit compatibility note in the docs.

PR #1626, opened 2026-08-24
feat(io): unify undecoding behavior and support overlong images list

Moenupa:
## Summary

- [x] (feat) add function `undecode_media` in `io_utils`, with a no-op bypass when nothing to do. It takes an optional batch_size which defaults to env var `DATASET_TF_BATCH_SIZE` or 1000.
- [x] (break) migrate `parquet_to_list` to use `undecode_media` for binary media
- [x] (break) migrate remote loading `RemoteLoader` to use `undecode_media` for binary media
- [x] (test) add tests to: 
  - [x] 1. non-media columns will maintain and keeps origianl values. 
  - [x] 2. regression tests to show why supporting `batch_size`: default `1000` may fail in extreme cases.
- [x] (doc) update env var docs.

## Tests

```sh
pytest tests/benchmark/test_general_vqa_adapter.py
pytest tests/benchmark/test_general_vmcq_adapter.py
pytest tests/api/test_io_utils.py
```

### DIFF
--- a/docs/en/get_started/parameters.md
+++ b/docs/en/get_started/parameters.md
@@ -12,7 +12,7 @@ The following environment variables can be set before launch to control global d
 | `EVALSCOPE_LANGUAGE` | Global default language, affects output language for reports, etc. (`en` or `zh`) | `en` |
 | `EVALSCOPE_HEARTBEAT_INTERVAL` | Heartbeat reporting interval (seconds) | `60` |
 | `MODELSCOPE_CACHE` | Root cache directory for ModelScope models and datasets | `~/.cache/modelscope/hub` |
-| `DATASET_TF_BATCH_SIZE` | Batch size for dataset transformation | `1000` |
+| `DATASET_TF_BATCH_SIZE` | Batch size for dataset transformation | `100` |
 
 ## Model Parameters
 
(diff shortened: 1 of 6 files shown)


Yunnglin:
Thanks for the improvement. Before merging, could you please validate the behavior for datasets without media columns?

`undecode_media()` currently always calls `dataset.cast(...)`, even when no feature is changed. Since `Dataset.cast()` internally runs a full `Dataset.map()`, this may add an unnecessary full pass and cache write for every text-only dataset loaded by `RemoteDataLoader`.

Could you please:
1. Verify the performance impact on a text-only dataset and return the original dataset when no media feature is found.
2. Add regression tests covering both a text-only dataset and the overlong image-list case.
3. Provide a before/after reproduction for the reported Arrow offset overflow and clarify why `batch_size=100` is appropriate.

Also, please add parameter and return type annotations to `undecode_media()`.

Moenupa:
Changes since the last review:
1. Restore default batch size to 1000; it also accepts a new env var `DATASET_TF_BATCH_SIZE`
2. Add a no-op bypass, if no media columns, skip casting return directly.
3. Add tests.

A small note:
Before this PR, casting will be done once per detected media column.
After this PR, casting will only be done once per dataset (which changes every media column).

Moenupa:
@Yunnglin I suggest we default batch size to 1000, for two good reasons:

1. Hugging Face default is 1000, going against a usual default requires special attention. 
   Plus, when authors are creating a dataset, they would have to handle this 1000 error.
   This is because hf `dataset.map()`, `dataset.filter()` all use this 1000 as batchsize.
2. This overflow hurts data transform speed and only occurs in very extreme cases. As a tradeoff we should consider the general-case scenario: 100+ images is extremely rare.

Yunnglin (approved):
LGTM. Verified the media undecoding changes locally, including the no-op path, Arrow offset overflow regression, and Image/Audio/Video sequence coverage. All targeted tests and CI checks pass.

PR #1628, opened 2026-08-25
fix(math/parser): use correct tag length when slicing bmatrix matrices

Ricardo-M-L:
Fix off-by-one in math_equal(): slicing always used pmatrix length for both pmatrix and bmatrix, producing corrupted content for bmatrix inputs.


Yunnglin:
Thanks for the contribution. After reviewing the change, the reported `bmatrix` off-by-one issue does not actually occur: `\begin{pmatrix}` and `\begin{bmatrix}` have the same length (15), while `\end{pmatrix}` and `\end{bmatrix}` also have the same length (13). Therefore, the proposed parser change is behaviorally equivalent to the existing implementation.

The `self.datset` typo fix has already been merged separately in #1589. In addition, the current pre-commit check is failing because YAPF reformats the parser change.

Since this PR does not introduce an effective new fix, we are closing it. Thanks again for taking the time to investigate and contribute.

PR #1630, opened 2026-08-25
fix(math/parser): use correct tag length when slicing bmatrix matrices

Ricardo-M-L:
Fix off-by-one in math_equal(): slicing always used pmatrix length for both pmatrix and bmatrix, producing corrupted content for bmatrix inputs.


Yunnglin:
Thanks for resubmitting this as a clean PR. However, the core issue remains unchanged: `\begin{pmatrix}` and `\begin{bmatrix}` are both 15 characters long, while `\end{pmatrix}` and `\end{bmatrix}` are both 13 characters long. Therefore, the old and new slicing expressions use identical indices and produce identical behavior. We also verified this with pmatrix/bmatrix combinations and found no behavioral difference. Since this does not provide an effective bug fix, we are closing the PR.

PR #1633, opened 2026-08-25
fix(score): prune rouge scoring function and fix en version of rouge

Moenupa:
## Summary

- [x] (break) remove unused, over-complicated `compute_rouge_score()`
- [x] (fix) `compute_rouge_score_one_sample()` had inverted reference and prediction.
- [x] (fix) add safeguards before Rouge functions, preventing strings from going into `zip()`
- [x] (feat) add a toggle to strictly match the lengths of `ref: list[str]` and `pred: list[str]` defaults to False, the same behavior as before.
- [x] no tests needed.
- [x] no doc updated needed.


Yunnglin (approved):
LGTM. Verified the ROUGE argument-order fix, input safeguards, type annotations, and cleanup. The failing unit test is unrelated to this PR (IQuiz dataset cache/download failure).

PR #1637, opened 2026-08-26
feat(benchmarks): add hle_tools benchmark

haoruilee:
## Summary

Add one new benchmark, the tool-use counterpart of closed-book `hle`:

- **Humanity's-Last-Exam-with-Tools** (`hle_tools`, dataset id `cais/hle`) — same 2,500 CAIS / Scale AI questions as `hle`, evaluated through EvalScope Native AgentLoop so the model can call tools before answering. This is a distinct leaderboard row from closed-book `hle`.

Dataset stays on ModelScope (`cais/hle`), same source as `hle`.

## Changes

- Add thin `HLEToolsAdapter` subclassing `HLEAdapter` (same judge, subsets, `include_multi_modal`, scoring)
- Default Native AgentLoop when `agent_config` is omitted: `function_calling`, `python_exec`, `local`, `max_steps=30`; optional MCP `fetch` if `evalscope[mcp]` is installed
- Add `test_hle_tools` next to `test_hle` (mock-safe local loop, no docker/network)
- Generate bilingual docs (EN/ZH) and `_meta` JSON
- Changelog entries in `README.md` / `README_zh.md`

## Test plan

- [x] Register `hle_tools` with `dataset_id=cais/hle`
- [x] Load 1 sample from ModelScope (`Other`, `include_multi_modal=False`)
- [x] Mock AgentLoop + HLE judge path
- [x] Live 1-sample eval on `deepseek-chat` via official API: pipeline completed (Accuracy 0% on that item — model miss, not a harness failure)
- [ ] CI `test_hle_tools`

## Notes

`datasets=['hle']` remains closed-book. `datasets=['hle_tools']` enables tools by default; passing `NativeAgentConfig` only overrides fields you set. Default environment is `local` for smoke tests; use `environment='docker'` for isolated formal runs. Web search is optional via MCP, not required for the default `python_exec` path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ For details, please refer to [📖 Visualizing Evaluation Results](https://evals
 
 ## 🎉 What's New
 
+- 🔥 **[2026.08.26]** Added **Humanity's Last Exam with Tools** (`hle_tools`) benchmark.
 - 🔥 **[2026.08.24] v1.11.0** Introduced published evaluation versions for reproducible benchmark results; improved report semantics and incomplete-run handling; strengthened multimodal media loading and task-config validation.
 - 🔥 **[2026.08.13]** Improved evaluation reports with unified metric semantics and more reliable Agent Trace step grouping and tool-call/result linking.
 - 🔥 **[2026.08.10]** Added **AutomationBench**, **JobBench**, **MiniWoB**, **OmniDocBench-v1.6**, **PerceptionBench**, **ScreenSpot-Pro**, **PLawBench**, **PMC-VQA**, **HiPhO**, **LogicVista**, and **CC-OCR-V2** benchmarks.
--- a/README_zh.md
+++ b/README_zh.md
@@ -75,6 +75,7 @@ EvalScope 提供交互式 Web Dashboard，支持多维度模型对比和深入�
 
 ## 🎉 内容更新
 
+- 🔥 **[2026.08.26]** 新增 **Humanity's Last Exam with Tools**（`hle_tools`）基准。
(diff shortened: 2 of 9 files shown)


Yunnglin:
Thanks for the contribution and for exploring tool-enabled HLE evaluation.

After reviewing the implementation, we do not think a separate `hle_tools` benchmark adapter is needed in its current form. EvalScope already supports running any `DefaultDataAdapter` benchmark, including `hle`, through the Native AgentLoop by configuring `TaskConfig.agent_config`, with registered tools or custom MCP servers.

The new adapter therefore mostly duplicates `hle` while hard-coding one particular AgentLoop configuration. Moreover, this configuration does not match the published HLE-with-tools protocol: the available tools, iteration limit, execution semantics, and completion protocol differ, so the resulting scores would not be directly comparable to existing HLE-with-tools results.

There are also some implementation concerns:

- `python_exec` defaults to the unisolated `local` environment, allowing model-generated code to execute directly on the host.
- The advertised sample tools can become inconsistent with the registered handlers when users override `agent_config.tools`.
- The added test does not produce or assert an actual tool call and is not currently exercised by the PR CI workflow.

For custom tool-assisted evaluation, users can already run `hle` with `NativeAgentConfig`. A documentation example or reusable configuration preset would be more appropriate and would avoid introducing another benchmark identity.

If the goal is to provide an official, independently reportable HLE-with-tools benchmark, we would need a new implementation that closely follows the published protocol, uses a safe execution environment by default, records the effective tool configuration for reproducibility, and includes scripted end-to-end tool-call tests.

For these reasons, we are closing this PR. Thanks again for the contribution and investigation.

PR #1651, opened 2026-08-27
fix(models): correct retry semantics and Anthropic streaming

git-jxj:
## Summary

This PR fixes three retry issues in model API calls:

- `retries=0` returned `None` without invoking the wrapped call, while `retries=None` raised `TypeError`.
- Non-recoverable client errors were retried with the normal delay before reaching backend error handling.
- Anthropic streaming requests were only protected while creating the stream; failures while consuming it escaped without retrying.

The changes cover synchronous and asynchronous paths across the affected OpenAI-compatible, OpenAI Responses, Anthropic, and LiteLLM backends.

## Root Cause

`retry_call()` and `async_retry_call()` iterated directly over `range(retries)`. Zero or negative values produced no attempts, and `None` was not accepted by `range()`.

The helpers caught every `Exception`, including deterministic 400/401/403/404/422 client failures. Backend-specific handlers only saw those exceptions after all configured attempts and delays had been exhausted.

For Anthropic streaming, `messages.create()` ran inside the retry helper but `collect_stream_response()` ran afterward. Once `create()` returned a stream object, an interrupted stream was outside the retry boundary.

## Changes

- Treat `retries` as the total attempt count, default `None` to 3, and clamp values below 1 to one attempt.
- Add `no_retry_exceptions` support to synchronous and asynchronous retry helpers.
- Bypass retries for non-recoverable 400/401/403/404/422 SDK errors.
- Keep rate limits, conflicts, timeouts, connection failures, and server errors retryable.
- Apply the exception policy to OpenAI-compatible, OpenAI Responses, Anthropic, and LiteLLM call sites.
- Move Anthropic stream creation and complete stream consumption into one retryable operation.
- Discard partial output from failed Anthropic attempts before issuing a fresh request.
- Add synchronous and asynchronous regression tests for helper semantics, backend exception policies, and interrupted Anthropic streams.

## Reproduction

```text
retry_call(lambda: "x", retries=0)    -> None
retry_call(lambda: "x", retries=None) -> TypeError
BadRequest with retries=5              -> 5 attempts
Authentication/NotFound, retries=5     -> 5 attempts
Anthropic interrupted stream           -> ConnectionError after attempt 1
```

The new regression suite reports 15 failures against the old source, including all three behaviors above.

After this change:

```text
retries=0                              -> one call, returns "x"
retries=None                           -> uses the default attempt count
non-recoverable client error           -> one attempt
transient connection error             -> retries to the configured attempt count
Anthropic interrupted stream           -> attempt 2 returns the complete response
```

## Validation

Targeted regression tests:

```bash
python -m pytest \
  tests/test_function_utils.py \
  tests/models/test_no_retry_exceptions.py \
  tests/models/test_anthropic_stream_retry.py -q
```

Result:

```text
58 passed
```

Model-layer regression suite:

```bash
```

Result:

```text
78 passed
```

CI smoke test:

```bash
```

Result:

```text
1 passed
```

All configured pre-commit checks also passed:

```text
flake8, ruff check, yapf, trailing whitespace, YAML, end-of-file,
```

## Scope

This change does not alter request payloads, configured retry intervals, or the number of attempts for values greater than zero. Rate limits and transient transport/server failures remain retryable. The new stream-consumption retry boundary is limited to Anthropic; unrelated backend streaming behavior is unchanged.

### DIFF
--- a/evalscope/models/anthropic_compatible.py
+++ b/evalscope/models/anthropic_compatible.py
@@ -1,8 +1,17 @@
 import os
 import time
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
-from anthropic import Anthropic, APIStatusError, AsyncAnthropic, BadRequestError, PermissionDeniedError
+from anthropic import (
+    Anthropic,
+    APIStatusError,
+    AsyncAnthropic,
+    AuthenticationError,
+    BadRequestError,
+    NotFoundError,
+    PermissionDeniedError,
+    UnprocessableEntityError,
+)
 from anthropic.types import Message
 
 from evalscope.api.messages import ChatMessage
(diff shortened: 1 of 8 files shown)


Yunnglin (approved):
LGTM!

PR #1653, opened 2026-08-27
fix(models): correct cache, usage, streaming, and image outputs

git-jxj:
## Summary

This PR fixes six independent model-layer correctness issues:

- Model memoization could return an instance created for a different `eval_type`.
- ModelScope generation with multiple choices only reported usage from one choice.
- Packed streaming choices could lose their individual `finish_reason` values.
- Text-to-image and image-editing outputs omitted the `data:` URI prefix required by downstream consumers.
- LiteLLM requests ignored `reasoning_history` and the configured base URL.
- Image model outputs stored an epoch timestamp instead of elapsed generation time.

Each behavior is covered by deterministic tests with mocked model or SDK dependencies.

## Root Cause

The model cache key did not include `eval_type`, so otherwise identical task and judge configurations could collide across backends.

ModelScope usage and OpenAI-compatible stream metadata were assigned from a single choice instead of being aggregated or indexed per choice.

Image model adapters serialized raw base64 content and assigned `time.time()` directly to `ModelOutput.time`, while downstream APIs expect a data URI and elapsed seconds.

The LiteLLM request builder did not forward the same reasoning-history and base-URL settings used by the OpenAI-compatible path.

## Changes

- Include `eval_type` in the model memoization key and document the cache contract.
- Aggregate ModelScope token usage across every generated choice.
- Track streaming `finish_reason` by choice index.
- Emit complete image data URIs from text-to-image and image-editing models.
- Report elapsed generation time for image outputs.
- Forward LiteLLM `reasoning_history` and `base_url` configuration.
- Add regression tests for all six behaviors.

## Reproduction

On the unmodified base:

```text
same model/config with different eval_type -> the same cached backend instance
ModelScope two-choice usage              -> 10 instead of 20
second packed choice finish_reason       -> stop instead of length
image output consumed as media           -> FileNotFoundError
image ModelOutput.time                   -> approximately 1.7e9 seconds
LiteLLM reasoning/base URL settings      -> silently omitted
```

The regression suite reports 11 failures against the old source. After this change, each configuration produces the intended backend, metadata, image payload, and timing value.

## Validation

Targeted regression tests:

```bash
python -m pytest \
  tests/models/test_model_cache.py \
  tests/models/test_modelscope_usage.py \
  tests/models/test_openai_stream_finish_reason.py \
  tests/models/test_image_model_output.py \
  tests/models/test_litellm_reasoning_history.py -q
```

Result:

```text
13 passed
```

Model-layer regression suite:

```bash
```

Result:

```text
62 passed
```

CI smoke test:

```bash
python -m pytest tests/cli/test_all.py::TestRun::test_ci_lite -q -p no:warnings
```

Result:

```text
1 passed
```

All current Ruff and repository pre-commit checks passed.

## Scope

These changes do not alter prompts, generation parameters, cache lifetime, retry behavior, image encoding content, or non-LiteLLM request routing. The branch is rebased onto the current `main` and contains one commit.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -156,7 +156,7 @@ An adapter must **never** call `self.llm_judge.judge()` or parse a judge reply i
 - `repeats`: duplicates items for k-metrics. `generation_config.n` is deprecated and mapped.
 - Use `generation_config` for runtime params. `TaskConfig.timeout` / `stream` are deprecated — forwarded with a warning.
 - `dataset_args` merges into `BenchmarkMeta._update()` (supports `local_path`, `filters` OrderedDict prepended).
-- Models are memoized by `(name, config, base_url, api_key, args)`.
+- Models are memoized by `(name, eval_type, config, base_url, api_key, args)`.
 - Use `@thread_safe` for model creation, `run_in_threads_with_progress` for concurrent eval.
 - Outputs land in `outputs/<timestamp>/{logs,predictions,reviews,reports,configs}/` (see `OutputsStructure`). `use_cache` resumes runs; `rerun_review` recomputes scores only.
 - `evalscope app` CLI command is **deprecated** (see `evalscope/cli/start_app.py`) — use `evalscope service` for the Web dashboard.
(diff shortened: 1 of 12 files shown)


Yunnglin (approved):
LGTM!

PR #1656, opened 2026-08-28
fix(benchmark): harden Toolathlon job lifecycle and result validation

git-jxj:
Hardens the Toolathlon private-mode client:

- Retry transient status, task archive, and result download failures without cancelling the remote job.
- Keep failed archive names pending locally in case the server no longer returns them.
- Recognize the official timeout state, task_names response field, and average_success_rate score.
- Handle relay exit and process cleanup without masking the original failure.
- Pass the local API key through the relay environment instead of process arguments.
- Reject missing or invalid result scores instead of treating them as zero.

The scoring and lifecycle changes move the benchmark evaluation version to v1.1.


Yunnglin:
Thanks for the thorough lifecycle hardening and regression coverage. The clean-relay-exit race fix, moving the API key out of argv, and reaping the relay process are all useful improvements.

I compared this implementation with the current official Toolathlon client/server code and found a few protocol and lifecycle differences that should be addressed before merge:

1. **Transient polling/download errors should not cancel the remote evaluation.**

   The official client treats incremental download and status-poll failures as recoverable: it logs them and continues polling. In this PR, any exception from `_download_completed_tasks()`, `_get_job_status()`, or result downloading reaches the outer handler and immediately calls `_cancel_job_safely()`.

   Consequently, a temporary network error, one HTTP 500, or a task archive download failure can cancel a multi-hour remote evaluation. Please keep polling/retrying transient service and download failures, and reserve cancellation for explicit local timeout, user interruption, relay failure that cannot recover, or another genuinely fatal condition.

2. **Please include the official `timeout` terminal state.**

   The official status poller handles `failed`, `timeout`, and `cancelled` as terminal failures. `_FAILED_JOB_STATES` currently omits `timeout`, so a remotely timed-out job may be misreported as an unexpected relay exit and cancelled again, or continue polling until the local timeout. Please add `timeout` and cover it with a regression test that verifies no second cancellation occurs.

3. **Please align completed-task parsing with the official response schema.**

   The official `/get_completed_tasks` endpoint returns:

   ```json
   {"task_names": [...]}
   ```

   but `_download_completed_tasks()` currently reads only `completed_tasks` or `tasks`. This can silently disable incremental task-result downloads. Please support `task_names` and update the mock flow to exercise the official field.

4. **Please recognize the official aggregate score field and version the scoring change.**

   The official `eval_stats.json` generator publishes the aggregate score as `average_success_rate`. `_extract_accuracy()` does not currently recognize it and instead depends on `eval_res_all.jsonl` as a fallback. Please support `average_success_rate`.

   Also, changing missing/invalid score handling from a silent `0.0` to an exception changes evaluation semantics. Per EvalScope's evaluation-versioning policy, Toolathlon should explicitly declare `evaluation_version='v1.1'`.

With these changes, the PR would preserve the useful hardening while remaining compatible with the official Toolathlon protocol and its retry behavior.

git-jxj:
Updated in fee18a08. Transient poll and download errors now retry without cancelling the job. Also added timeout, task_names, average_success_rate, and v1.1 handling. Failed archive downloads remain pending locally so they can be retried if the server no longer returns the task name.

Yunnglin (approved):
LGTM. The lifecycle hardening, protocol compatibility updates, score validation, and generated Toolathlon documentation are consistent and well covered. Thanks for addressing the review feedback.

PR #1657, opened 2026-08-28
fix(cache): deduplicate review state and tolerate torn resume rows

git-jxj:
## Summary

This PR makes `use_cache` recovery resilient to stale, duplicated, and partially written JSONL cache rows.

- Review-cache rows now intersect with the sample IDs in the current task and deduplicate by sample ID, keeping the last appended row.
- Corrupt JSONL, `null`, and other non-object rows are skipped only when callers explicitly opt into cache-tolerant loading.
- Prediction and review cache rows that fail their Pydantic schema are skipped with warnings instead of aborting the entire resume.

## Root Cause

`CacheManager.filter_review_cache()` deserialized every row in the review JSONL and returned each resulting `SampleScore`. It used the IDs only to remove pending task states, so stale rows from a changed dataset and duplicate rows for a sample were still included in aggregation.

`jsonl_to_list()` retried parsing line by line after the `jsonlines` reader failed, but the fallback called `json.loads()` on every non-empty row without handling malformed records. A crash-truncated tail therefore failed the fallback as well. A literal `null` row could also reach cache model validation as `None`.

## Changes

- Keep cached review scores in a map keyed by `sample_id`, accept only IDs in the current task, and retain the last row for duplicates.
- Log dropped orphan and duplicate review rows so cache recovery is observable.
- Add an explicit `skip_invalid` mode to `jsonl_to_list`; its default remains strict for benchmark and other normal JSONL inputs.
- In tolerant mode, parse JSONL line by line and skip malformed and non-object records with a warning.
- Make prediction and review cache restoration skip individual `ModelResult` or `ReviewResult` schema failures.
- Add regression coverage for orphan rows, last-row-wins deduplication, repeat samples, missing and partial caches, malformed tails, `null`/non-object JSON, and invalid cache schemas.

## Reproduction

On the unmodified base, review states `0, 1` and the following cache rows:

```text
0/1.0, 1/0.0, 1/1.0, 99/0.5
```

return four scores (`[0, 1, 1, 99]`) and include the stale row in metrics. The correct result is two scores (`[0, 1]`) with the final value for sample `1`.

Likewise, reading a cache ending with a truncated JSON object raises `JSONDecodeError`, and a valid `null` line subsequently raises `ValidationError` during review restoration. Either failure prevents `use_cache` from resuming an interrupted evaluation.

## Validation

```bash
python -m pytest \
  tests/evaluator/test_cache_manager.py \
  tests/api/test_io_utils.py -q
```

Result:

```text
21 passed
```

```bash
```

Result:

```text
1 passed
```

All configured pre-commit hooks passed when run through the project virtual environment. `make lint` could not locate the shell `pre-commit` executable in this environment, but the equivalent `python -m pre_commit run --all-files` completed successfully.

## Scope

The change is limited to cache restoration and its JSONL reader contract. Normal JSONL reads remain strict, live samples are not deduplicated across distinct IDs, and no benchmark prompts, scoring definitions, aggregation rules, or published `evaluation_version` values change.

### DIFF
--- a/evalscope/api/evaluator/cache.py
+++ b/evalscope/api/evaluator/cache.py
@@ -3,7 +3,7 @@ import os
 import uuid
 from typing import Any, Dict, List, Optional, Tuple
 
-from pydantic import BaseModel, ConfigDict, Field, model_validator
+from pydantic import BaseModel, ConfigDict, Field, ValidationError, model_validator
 
 from evalscope.api.agent import AgentTrace
 from evalscope.api.dataset import Dataset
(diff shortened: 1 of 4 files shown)


Yunnglin (approved):
LGTM! The cache recovery changes are focused and well tested. Invalid, stale, and duplicate cache rows are handled safely while strict JSONL behavior remains the default. Targeted tests and all required CI checks pass.

PR #1659, opened 2026-08-28
fix(dataset): isolate cache keys, undecode video, and validate limits consistently

git-jxj:
Fixes three native dataset-loading issues:

- Include the dataset hub, split, subset, version, and loader arguments in the remote cache identity to avoid collisions.
- Keep remote video columns undecoded when materializing datasets.
- Validate limit consistently in TaskConfig, DatasetDict, and direct data loaders without mutating loader state across repeated loads.

Also keeps local-path precedence consistent and adds regression coverage for cache isolation, video loading, invalid limits, and repeated loads.

### DIFF
--- a/evalscope/api/dataset/dataset.py
+++ b/evalscope/api/dataset/dataset.py
@@ -1,6 +1,7 @@
 import abc
 import copy
 import hashlib
+import math
 import random
 from collections import defaultdict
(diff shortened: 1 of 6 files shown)


Yunnglin (approved):
LGTM

PR #1662, opened 2026-08-28
fix(models): avoid blocking event loop during async audio preprocessing

git-jxj:
## Summary

This PR prevents HTTP audio preprocessing from blocking the event loop in OpenAI-compatible asynchronous generation.

- Audio URLs are still downloaded and converted to the Base64 form required by the OpenAI-compatible API.
- Synchronous preprocessing runs outside the async event-loop thread.
- Extensionless audio sources now use their declared audio MIME type when a data URI is required.

## Root Cause

`generate_async()` constructed chat messages synchronously before its first await. For an HTTP audio URL,
`file_as_data_uri()` calls `requests.get(..., timeout=30)`, so one slow download blocked every task sharing the event loop.

Passing the URL directly as `input_audio.data` is invalid: OpenAI-compatible audio input requires Base64 data. The fix preserves the existing wire format and changes only where preprocessing executes.

## Changes

- Build OpenAI-compatible messages through `asyncio.to_thread()` in the asynchronous generation path.
- Keep HTTP audio URL downloads encoded as Base64 for standard OpenAI/vLLM endpoints.
- Preserve DashScope data-URI behavior and use `audio/wav` or `audio/mpeg` as the fallback MIME type.
- Add mocked regression coverage for event-loop responsiveness, payload encoding, provider behavior, and download failures.

## Reproduction

On the unmodified base, an async heartbeat scheduled alongside `generate_async()` is delayed by a mocked 200 ms audio download because the blocking request runs before the first await. After this change, the heartbeat continues while preprocessing runs in a worker thread, and the outgoing audio payload remains valid Base64.

## Validation

```bash
python -m pytest tests/models/test_openai_audio_input.py \
  tests/models/test_openai_reasoning.py \
  tests/models/test_openai_stream_reasoning.py \
  tests/models/test_openai_stream_retry.py \
  tests/models/test_async_client_lifecycle.py \
  tests/api/test_seed_tts_eval_benchmark.py \
  tests/api/test_text2speech_model.py -q
# 42 passed

# 1 passed

make lint
# passed (using the repository virtual environment)
```

All downloads are mocked; no external model or audio service is required.

## Scope

The change is limited to OpenAI-compatible audio preprocessing and its regression tests. It does not send raw URLs in `input_audio.data`, alter synchronous generation, or change reasoning, tool-call, or streaming request structure.

### DIFF
--- a/evalscope/models/openai_compatible.py
+++ b/evalscope/models/openai_compatible.py
@@ -1,3 +1,4 @@
+import asyncio
 import os
 import time
 from typing import Any, Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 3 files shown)


Yunnglin (approved):
LGTM!

PR #1663, opened 2026-08-28
fix(resources): verify NLTK mirror archives with pinned digests

git-jxj:
## Summary

This PR makes fallback NLTK resource downloads fail closed unless the downloaded archive matches a pinned SHA-256 digest.

- Pin a digest for every enabled punkt_tab, stopwords, and averaged_perceptron_tagger_eng mirror URL.
- Pass the URL-specific digest to the existing download helper before extraction.
- Remove the Gitee punkt_tab fallback because it has no trusted digest.
- Add offline regression coverage for digest configuration, digest forwarding, checksum failure, and cleanup.

## Root Cause

When a required NLTK resource was absent, check_nltk_data downloaded a ZIP from a mutable third-party mirror and extracted it directly. Although download_url already supports checksum verification, the call did not provide sha256. A valid but replaced or corrupted archive could therefore be installed into the NLTK data directory.

## Changes

- Move the NLTK mirror configuration to a module-level MIRROR_MAP with per-URL SHA-256 values.
- Use download_url(url, zip_path, sha256=...) for every enabled mirror.
- Keep extraction after the verified download only; a checksum failure raises and does not invoke ZipFile.
- Verify the three pinned archives again through the configured proxy before this commit:
  - ModelScope punkt_tab: c2b16c23d738effbdc5789d7aa601397c13ba2819bf922fb904687f3f16657ed
  - Gitee stopwords: 15c94179887425ca1bedc265608cab9f27d650211f709bb929e320990a4b01d1
  - NLTK GitHub averaged_perceptron_tagger_eng: 6025f530624335c67d6547d44757b357b4e79bae030a0383e9887a92c1718f0b

## Reproduction

On the unmodified base, monkeypatching download_url to record its keyword arguments during check_nltk_data(averaged_perceptron_tagger_eng) records sha256 as not passed. The downloader then skips the checksum comparison and resource_utils extracts the returned archive.

## Validation

Targeted resource tests:

```text
python -m pytest tests/test_resource_utils.py -q
6 passed
```

Adjacent NLTK consumer tests:

```text
python -m pytest tests/benchmark/test_ifeval_language_checkers.py -q
2 passed
```

CI smoke test:

```text
python -m pytest tests/cli/test_all.py::TestRun::test_ci_lite -q -s -p no:warnings
1 passed
```

Lint:

```text
PATH=/root/code/evalscope/.venv/bin:$PATH make lint
Passed
```

All added tests use mocked NLTK and download behavior; no test accesses the network.

## Scope

The change is limited to NLTK mirror integrity and its tests. It intentionally reuses the existing checksum-capable download helper and does not include HTTP resume or other download behavior changes.

### DIFF
--- a/evalscope/utils/resource_utils.py
+++ b/evalscope/utils/resource_utils.py
@@ -16,6 +16,45 @@ logger = get_logger()
 # Path to the benchmark metadata directory (individual JSON files per benchmark)
 BENCHMARK_META_DIR = Path(__file__).parent.parent / 'benchmarks' / '_meta'
 
+# Mirror archives are pinned per URL because mirrors can serve different
+# nltk_data snapshots. Only archives with a verified digest may be extracted.
+MIRROR_MAP: Dict[str, Dict[str, Any]] = {
+    'punkt_tab': {
+        'resource_path': 'tokenizers/punkt_tab',
+        'zip_name': 'punkt_tab.zip',
+        'extract_dir': 'tokenizers',
+        'mirrors': [
+            {
+                'url': 'https://modelscope-open.oss-cn-hangzhou.aliyuncs.com/open_data/nltk_data/punkt_tab.zip',
+                'sha256': 'c2b16c23d738effbdc5789d7aa601397c13ba2819bf922fb904687f3f16657ed',
+            },
+        ],
+    },
+    'stopwords': {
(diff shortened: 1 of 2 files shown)


Yunnglin (approved):
LGTM!

PR #1664, opened 2026-08-28
fix(judge): serialize lazy judge initialization

git-jxj:
## Summary

This PR makes lazy LLM Judge construction safe when an evaluation starts reviewing samples concurrently.

- A benchmark now initializes its configured judge list exactly once, even when multiple review workers reach the first score simultaneously.
- All concurrent callers receive the same initialized list and judge instances.
- A failed initialization leaves no partial result published, so a later access can retry.
- Rule-only scoring continues to return an empty list without constructing a judge.

## Root Cause

`LLMJudgeMixin.llm_judges` used an unlocked check-then-initialize sequence. With LLM Judge scoring and `eval_batch_size > 1`, every worker could observe `_llm_judges is None` before any assignment completed, then each construct its own judge list. The evaluator executes per-sample review in a thread pool, so this occurs on the first concurrent review.

## Changes

- Add a dedicated `_llm_judges_lock` to `LLMJudgeMixin`.
- Use double-checked locking around lazy judge construction.
- Keep the executor lock separate: executor construction reads `llm_judges`, so sharing the lock could introduce nested-lock deadlock risk.
- Add deterministic barrier-based regression coverage for concurrent construction, object identity, failure-and-retry behavior, and disabled judge scoring.

## Reproduction

On the unmodified base, eight worker threads synchronized at their first `llm_judges` access each enter `init_llm_judges()`:

```text
init_calls = 8
unique judge-list instances = 8
```

After this change:

```text
init_calls = 1
all workers receive the same judge list and judge instance
```

## Validation

```bash
python -m pytest \
  tests/api/judge/test_llm_judge_mixin.py \
  tests/api/judge/test_migrated_adapters.py \
  tests/api/judge/test_executor.py \
  tests/api/judge/test_contracts.py -q
```

Result: `75 passed`.

```bash
```

Result: `1 passed`.

All configured repository pre-commit hooks passed via the project virtual environment. `make lint` could not find the shell `pre-commit` executable in this environment, but the equivalent `python -m pre_commit run --all-files` completed successfully.

## Scope

Changes are limited to `LLMJudgeMixin` lazy initialization and regression tests. Judge configuration, scoring semantics, the dynamic judge setter contract, benchmark metadata, and published `evaluation_version` values are unchanged. The branch is rebased onto current `main` and contains one commit.

### DIFF
--- a/evalscope/api/mixin/llm_judge_mixin.py
+++ b/evalscope/api/mixin/llm_judge_mixin.py
@@ -66,6 +66,7 @@ class LLMJudgeMixin:
         self._task_config = task_config
 
         self._llm_judges: Optional[List[LLMJudge]] = None
+        self._llm_judges_lock = threading.Lock()
         self._judge_executor: Optional['JudgeExecutor'] = None
         self._judge_executor_lock = threading.Lock()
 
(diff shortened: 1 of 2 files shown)


Yunnglin (approved):
LGTM!

PR #1673, opened 2026-08-28
fix(models): retry 200 responses that carry a gateway error payload

### LINKED ISSUE #1674: One transient gateway error aborts the whole run: choices=None raised outside the retry boundary
## Self-Check List

- [x] I have carefully read the [relevant user documentation](https://evalscope.readthedocs.io/en/latest/get_started/parameters.html)
- [x] I have reviewed the [Frequently Asked Questions](https://evalscope.readthedocs.io/en/latest/get_started/faq.html) — it does not cover this error
- [x] I have searched and reviewed existing issues to confirm this is not a duplicate problem

## Problem Description

While running GSM8K against the free-tier model `minimax/minimax-m3:free` on OpenRouter, some requests came back with `200 OK` — but the body wasn't a completion at all. It was an error saying the account balance had run out:

```json
{"error":{"message":"Insufficient balance","code":402}}
```

The OpenAI SDK parses that into a `ChatCompletion` with `choices=None`, so to everything downstream it still looks like a successful call.

**The response is checked in the wrong place.** [`retry_call`](https://github.com/modelscope/evalscope/blob/3c52bd0923b75b2a83b776d6f526feb70572bfdf/evalscope/models/openai_compatible.py#L156-L161) only covers the HTTP call, and nothing raises there — the request *succeeded*. The failure only surfaces later at [line 169](https://github.com/modelscope/evalscope/blob/3c52bd0923b75b2a83b776d6f526feb70572bfdf/evalscope/models/openai_compatible.py#L169), inside [`chat_choices_from_openai`](https://github.com/modelscope/evalscope/blob/3c52bd0923b75b2a83b776d6f526feb70572bfdf/evalscope/models/utils/openai.py#L673-L679), which is outside the retry block.

So `retries=5` can never apply, and with the default `ignore_errors=False` one bad response ends the whole run. Mine died at sample 44 of 1319.

The error message is also unhelpful — the real reason survives in `model_extra`, but nothing reads it, so you only see `id=None, model=None`.

## EvalScope Version (Required)

v1.11.0 (also reproduced on `main` @ `3c52bd0`)

## Tools Used

- [x] Native / Native framework

## Executed Code or Instructions

```python
import os
from evalscope import TaskConfig, run_task

OR_KEY = os.getenv("OPENROUTER_API_KEY")
OR_BASE = "https://openrouter.ai/api/v1"

MODEL = 'minimax/minimax-m3:free'

task_cfg = TaskConfig(
    model=MODEL,
    api_url=OR_BASE,
    api_key=OR_KEY,
    eval_type="openai_api",

    datasets=["gsm8k"],

    # limit=5,                  # quick sanity check
    eval_batch_size=5,

    generation_config={
        "temperature": 0,
        "max_tokens": 1024,
    },

    dataset_args={
        "gsm8k": {
            "few_shot_num": 0,
            "few_shot_random": False,
        }
    }
)

run_task(task_cfg)
```

It depends on the upstream provider failing mid-run, so it won't reproduce on every attempt. It happened at sample 44 of one run, and three times during a later 200-sample run.

## Error Log

```
Evaluating[gsm8k]:   3%|██▉  | 44/1319 [00:55<24:59,  1.18s/it]
ERROR: Model [minimax/minimax-m3:free] returned an invalid response: Model response
contains no choices (choices=None). Response id=None, model=None. This may indicate
the model returned an empty or malformed response.

  File "evalscope/evaluator/evaluator.py", line 515, in _predict_sample
    task_state = self.benchmark.run_inference(model=self.model, sample=sample, ...)
  File "evalscope/api/benchmark/adapters/default_data_adapter.py", line 437, in _on_inference
    model_output = model.generate(input=sample.input, tools=sample.tools)
  File "evalscope/models/openai_compatible.py", line 145, in generate
    choices = self.chat_choices_from_completion(completion, tools)
  File "evalscope/models/utils/openai.py", line 666, in chat_choices_from_openai
    raise ValueError(
ValueError: Model response contains no choices (choices=None). Response id=None,
model=None. This may indicate the model returned an empty or malformed response.
```

## Running Environment

- Operating System: Linux 6.6.87.2-microsoft-standard-WSL2
- Python Version: 3.12.3

## Additional Information

Fix in #1673: check the response where it comes back and raise a clear error if the body carries one, so the retry can act on it and the gateway's real message is reported. Only the `choices=None` case changes; real HTTP errors are untouched.

Workaround until then: set `ignore_errors=True`, which costs the affected sample instead of the run.

https://claude.ai/code/session_0195nPGwxPjJGtQukF28BJfS

seroze:
Closes #1674

## Summary

Some OpenAI-compatible gateways return an error payload with `200 OK`. The OpenAI SDK deserializes it as a `ChatCompletion` with `choices=None`, so the error previously surfaced outside EvalScope's retry boundary with limited diagnostics.

This PR:

- validates non-streaming completions inside the existing sync and async retry boundary;
- reports the gateway error preserved in `model_extra`;
- retries transient or unclassified gateway failures;
- fails fast for deterministic `400`, `401`, `402`, `403`, `404`, and `422` errors;
- preserves the existing `ValueError` compatibility and streaming behavior.

## Tests

Added deterministic, network-free coverage for:

- OpenAI SDK deserialization of gateway error payloads;
- sync and async transient-error retries;
- retry exhaustion and missing error details;
- non-retryable gateway client errors;
- valid completions without unnecessary retries.

### DIFF
--- a/evalscope/models/openai_compatible.py
+++ b/evalscope/models/openai_compatible.py
@@ -52,6 +52,32 @@ NON_RETRYABLE_OPENAI_ERRORS: Tuple[Type[Exception], ...] = (
 )
 
 
+class EmptyCompletionError(RuntimeError):
+    """A 200 response whose body carried no choices (usually a gateway error)."""
+
+
+def raise_if_error_payload(completion: ChatCompletion) -> None:
+    """Reject an error payload that the SDK deserialized as a completion.
+
+    Some gateways (OpenRouter among them) commit ``200 OK`` headers before the
+    upstream has produced anything, holding the connection open with whitespace
+    padding. When the request then fails, the status line is already sent, so
+    the error can only be appended to the body:
+
+        {"error": {"message": "Insufficient balance", "code": 402}}
+
+    The SDK parses that into a ``ChatCompletion`` whose ``choices`` is None and
+    keeps the real error in ``model_extra``. Raising here — inside the retry
+    boundary — turns a silent, run-killing corruption into a retried request
+    that reports what the gateway actually said.
+    """
+    if completion.choices is not None:
+        return
+    error = (completion.model_extra or {}).get('error')
+    detail = f': {error}' if error else ' (no error detail in body)'
+    raise EmptyCompletionError(f'Gateway returned a response with no choices{detail}')
+
+
 class OpenAICompatibleAPI(ModelAPI):
     def __init__(
         self,
@@ -150,6 +176,7 @@ class OpenAICompatibleAPI(ModelAPI):
             def _create_and_collect() -> Tuple[ChatCompletion, Optional[float]]:
                 raw_completion = self.client.chat.completions.create(**request)
                 if isinstance(raw_completion, ChatCompletion):
+                    raise_if_error_payload(raw_completion)
                     return raw_completion, None
                 return collect_stream_response(raw_completion, request_start=t_start)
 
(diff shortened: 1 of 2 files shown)


Yunnglin (approved):
LGTM

PR #1676, opened 2026-08-28
fix(ifbench): enforce unique words in sentence checker

### LINKED ISSUE #1675: [Bug]: IFBench ratio:sentence_words accepts repeated words
## Self-Check List

- [x] I have reviewed the relevant documentation and existing issues.
- [x] I confirmed this is not covered by an open or closed PR.

## Problem Description

`CharacterCountUniqueWordsChecker` describes the constraint as: "Respond with three sentences, all containing the same number of characters but using all different words." However, `check_following()` validates only the sentence count and equal character lengths. It never checks word uniqueness.

This creates deterministic false positives in IFBench strict and loose scores. For example, `Cat. Cat. Cat.` has three equal-length sentences and is currently accepted even though every word is repeated.

## Minimal Reproduction

```python
from evalscope.benchmarks.ifbench.instructions import CharacterCountUniqueWordsChecker

checker = CharacterCountUniqueWordsChecker("ratio:sentence_words")
assert checker.check_following("Cat. Cat. Cat.") is True  # current behavior
```

## Expected Behavior

The checker should reject a response when any case-insensitive word token is repeated across the three sentences, while preserving the existing sentence-count and character-count checks. Standalone punctuation tokens should not participate in uniqueness.

## Proposed Fix

Reuse the existing `_word_tokens_without_punctuation()` helper, normalize tokens with `casefold()`, and require the token count to equal the unique-token count. Add focused offline tests for repeated words, case-insensitive duplicates, punctuation, unequal sentence lengths, and a valid response.

## EvalScope Version

Current `main` at the time of filing.

## Tools Used

- [x] Native / Native framework

linhongyu510:
## Summary

- enforce the `ratio:sentence_words` requirement that all words across the three sentences are distinct
- compare tokens case-insensitively with `casefold()` while reusing the existing punctuation-filtering tokenizer
- reject punctuation-only responses that contain no valid word tokens
- preserve the existing exact-three-sentences and equal-character-count checks
- bump IFBench `evaluation_version` from `v1.0` to `v1.1` so cached scores cannot cross the scoring-semantics change

## Why

`CharacterCountUniqueWordsChecker` currently checks only the sentence count and character lengths. As a result, responses such as `Cat. Cat. Cat.` receive a false-positive IFBench score even though the instruction explicitly requires all different words.

## Tests

- `pytest -q tests/benchmark/test_ifbench_instructions.py` - 11 passed
- `pytest -q tests/benchmark/test_eval.py -k ifbench` - 2 passed, 130 deselected
- `pytest -q tests/api/test_evaluation_versioning.py` - 10 passed
- `pre-commit run --files evalscope/benchmarks/ifbench/instructions.py evalscope/benchmarks/ifbench/ifbench_adapter.py tests/benchmark/test_ifbench_instructions.py` - passed

Fixes #1675

### DIFF
--- a/evalscope/benchmarks/ifbench/ifbench_adapter.py
+++ b/evalscope/benchmarks/ifbench/ifbench_adapter.py
@@ -57,6 +57,7 @@ IFBench is a benchmark designed to evaluate how reliably AI models follow novel,
         few_shot_num=0,
         train_split=None,
         eval_split='train',
+        evaluation_version='v1.1',
     )
 )
 class IFBenchAdapter(DefaultDataAdapter):
--- a/evalscope/benchmarks/ifbench/instructions.py
+++ b/evalscope/benchmarks/ifbench/instructions.py
@@ -932,7 +932,8 @@ class CharacterCountUniqueWordsChecker(Instruction):
         for sentence in sentences:
             if len(sentence.strip()) != char_count:
                 return False
-        return True
+        words = [word.casefold() for word in _word_tokens_without_punctuation(value)]
+        return bool(words) and len(words) == len(set(words))
 
 
 class NthWordJapaneseChecker(Instruction):
--- a/tests/benchmark/test_ifbench_instructions.py
+++ b/tests/benchmark/test_ifbench_instructions.py
@@ -1,0 +1,59 @@
+import pytest
+
+from evalscope.api.registry import BENCHMARK_REGISTRY
+from evalscope.benchmarks.ifbench.ifbench_adapter import IFBenchAdapter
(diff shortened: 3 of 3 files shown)


linhongyu510:
The `Lint Test` and `CI Tests Lite` workflows are currently waiting with `action_required` for this fork PR. The branch is a single focused commit, is not behind `main`, and the changed implementation/test files pass Ruff and formatting checks locally. Could a maintainer approve the workflow runs when convenient?

Yunnglin (approved):
LGTM

PR #1679, opened 2026-08-29
feat: add OrcaRouter as a first-class evaluation provider

dulcestentaciones2920-debug:
Add `orcarouter_api` as a first-class eval-type in EvalScope's model API registry, so users of this one-stop LLM evaluation framework can evaluate through the [OrcaRouter](https://www.orcarouter.ai) gateway directly. OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents: like OpenRouter it exposes a provider/model namespace across many models, but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Making it a named provider means EvalScope users get that stack without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

This mirrors how `openai_api` / `anthropic_api` are wired today: a new `OrcaRouterCompatibleAPI` in `evalscope/models/` reuses the existing OpenAI-compatible chat path, registered via `@register_model_api` and added to `EvalType`, `REMOTE_API_EVAL_TYPES`, and the CLI `--eval-type` choices. The provider defaults to `https://api.orcarouter.ai/v1` and reads `ORCAROUTER_API_KEY`.

**Verification**
- `ruff check` + `ruff format` clean on all touched files
- `pytest tests/models/test_orcarouter_compatible.py tests/api/test_task_config_validation.py` — 47 passed
- Live-tested `OrcaRouterCompatibleAPI` against the real gateway (got a non-empty chat completion, 200)

I'm an engineer on the OrcaRouter team. Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter


Yunnglin:
Thank you for the contribution and for providing tests.

After reviewing the proposal, we do not plan to add a dedicated `eval_type` or documentation entry for OrcaRouter. OrcaRouter exposes an OpenAI-compatible API and is already fully supported through EvalScope's existing `openai_api` integration by supplying its base URL and API key.

This PR does not introduce a new protocol or EvalScope-specific capability; it mainly adds vendor-specific endpoint and environment-variable defaults. Adding dedicated aliases for individual OpenAI-compatible gateways would expand the public API and create an ongoing maintenance precedent for other providers.

For these reasons, we will not merge this PR. Users can continue to access OrcaRouter through the generic `openai_api` configuration. Thank you again for your interest in EvalScope and for taking the time to prepare the contribution.

PR #1689, opened 2026-09-01
chore: enforce F401 in evalscope/utils

Moenupa:
## Summary

- [x] remove unused imports in `evalscope/utils`
- [x] `__init__.py` now ignores rule F401 to prevent c
- [x] no doc update needed
- [x] no test update needed

## Tests

```sh
# regression test 1: checks for uses of undefined names.
# to ensure we dont remove used imports
ruff check --select=F821
```

## Criteria

ruff auto-fix is reliable if the removed import satisfies any of:
1. is imported from typing
2. is conditioned under `if TYPE_CHECKING:`
3. has no decorators (a counterexample would be registration of benchmarks, which must NOT be lazy loaded or non-imported).

Different from the *F401 is safe except `__init__.py`* claim in [ruff doc](https://docs.astral.sh/ruff/rules/unused-import/), decorators for registry requires import to tigger.

In our case, we have a few `register_*()` functions serving as decorators. Removing imports with these decorators will be unsafe.

### DIFF
--- a/evalscope/utils/data_utils.py
+++ b/evalscope/utils/data_utils.py
@@ -4,7 +4,7 @@ Data loading and processing utilities for reports and predictions.
 
 import glob
 import os
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List
 
 import pandas as pd
 from pydantic import ValidationError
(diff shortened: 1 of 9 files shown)


Yunnglin (approved):
LGTM

PR #1690, opened 2026-09-01
feat(tau3_bench): support configurable max_steps parameter

ZTJiu:
Pass max_steps to tau2.run_task() to allow control over the maximum number of steps the agent can execute per task.


Yunnglin (approved):
LGTM — `max_steps` is now documented in the generated benchmark pages and the English/Chinese τ³-bench guides. Verified the documentation generation, both HTML builds, Lint Test, and CI Tests Lite.

Issue #1631, opened 2026-08-25
Followup to #1629: CICD Improvements


Moenupa:
## Summary

#1629 proposes the migration to `ruff`. This issue document possible improvements that enhances the CICD workflow (mostly `pre-commit`).

Since CICD causes a lot of file updates, I will leave all changes to maintainers and only propose ways to improve. (I found issue > PR, due to a lot of conflicts and merges.)

## Linting

- [x] I ([isort](https://docs.astral.sh/ruff/rules/#isort-i))
- [x] E ([pydocstyle error](https://docs.astral.sh/ruff/rules/#pycodestyle-e-w))
- [x] W ([pydocstyle warning](https://docs.astral.sh/ruff/rules/#pycodestyle-e-w))
- [x] ~~B ([flake8-bugbear](https://docs.astral.sh/ruff/rules/#flake8-bugbear-b))~~
- [x] ~~Q ([flake8-quotes](https://docs.astral.sh/ruff/rules/#flake8-quotes-q))~~
- [ ] UP ([pyupgrade](https://docs.astral.sh/ruff/rules/#pyupgrade-up))
- [ ] SIM ([flake8-simplify](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim))
- [x] remove flake8 from pre-commit hook
- [ ] F401 in `evalscope`
- [ ] E/F/W rules in `tests`

## Formatting

- [x] add ruff-format and remove yapf in pre-commit hook (example: https://github.com/vllm-project/vllm/pull/26247)

## Why do we need migration?

In addition to discussion #1629, ruff brings a faster, unified lint+format experience, as flake8 and yapf may conflict, an example:

```py
match var:
    # some comments here
    case 1:
        pass
```

flake8 will always format as above, while yapf will always format as below. This makes `make lint` never pass.

```py
match var:
# some comments here
    case 1:
        pass
```

Moenupa:
@Yunnglin please consider reopening this issue.

I believe some of the rules in `ignore = ["E501", "E741", "F401", "F403", "F405", "F541"]` are useful and reduce runtime. I noticed a lot of unused imports and other issues picked up by these rules.

Also, we can gradually enforce lint/format in tests to standardize code style.

PR #1691, opened 2026-09-01
fix(cli): enforce F401 in cli and fix benchmark info import errors

Moenupa:
## Summary

- [x] CICD: enforce F401 in `evalscope/cli/` and `evalscope/report/` (#1631)
- [x] fix: benchmark-info raises import error when `--format markdown` is passed
- [x] typing: fix typing issues for `CLICommand` and its subclasses.
- [x] typing: add type hints to method in `BenchmarkInfoCMD`.
- [x] no doc update needed
- [x] no test update needed, CLI functions not affected, only type checks and hints.

## Test

```sh
evalscope benchmark-info mmlu --format markdown
```

### DIFF
--- a/evalscope/cli/base.py
+++ b/evalscope/cli/base.py
@@ -1,7 +1,10 @@
 # Copyright (c) Alibaba, Inc. and its affiliates.
 
 from abc import ABC, abstractmethod
-from argparse import ArgumentParser
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from argparse import ArgumentParser, _SubParsersAction
 
 
 class CLICommand(ABC):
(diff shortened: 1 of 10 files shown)


Yunnglin:
Thanks for the fix. I verified that `benchmark-info ... --format markdown` now works.

One issue before merging: `_SubParsersAction` is imported only under `TYPE_CHECKING` and used in a string annotation. On our supported Python 3.10, `typing.get_type_hints(CLICommand.define_args)` now raises `NameError: name '_SubParsersAction' is not defined`. Moreover, the runtime `argparse._SubParsersAction` is not subscriptable on Python 3.10.

Could we avoid exposing this private generic type in runtime-resolvable annotations? A local `Protocol` describing `add_parser()`, or a broader stable type, would avoid this compatibility regression. Please also add a small regression test for the markdown fallback path while updating this.

Moenupa:
My bad. I think `'_SubParsersAction'` is a higher-Python private. I was using 3.14 and did not check for 3.10.

I can revert only the `'_SubParsersAction'`-related stuff to make this PR cleaner.

Moenupa:
used protocol instead, as per suggested.

Yunnglin (approved):
LGTM

PR #1692, opened 2026-09-01
fix(dep): remove dependency `overrides`; type check instead of runtime

Moenupa:
## Summary

- [x] remove `overrides` dependency and use existing `typing_extension` instead: `@override` now checks during development type check, instead of runtime.
- [x] modify `evalscope/api/benchmark/adapters/default_data_adapter.py` to match `evalscope/models/modelscope.py`.
- [x] use torch.Tensor instead of Tensor, removing a `from import` to clean code.
- [x] no test update needed.
- [x] no doc update needed.


Yunnglin (approved):
LGTM

PR #1696, opened 2026-09-03
docs(agent): point prediction docstrings at the hook that resolves it

### LINKED ISSUE #1695: AgentLoop docstrings point at AgentTrace.final_prediction, which does not exist
## Summary

Three docstrings describe the reported prediction as living in `AgentTrace.final_prediction`. `AgentTrace` has no such field, and nothing in the repo ever sets it.

## Where

**`evalscope/api/agent/loop.py:363`** — `AgentLoop._emit_submit`

> ``final_answer`` here is what the *loop observed*, not the prediction that gets reported: that is resolved after the loop returns and lives in ``AgentTrace.final_prediction``.

**`evalscope/api/agent/loop.py:384`** — `AgentLoop._emit_implicit_submit`

> Only a short preview is kept, for diagnosis; the reported prediction is ``AgentTrace.final_prediction``.

**`tests/agent/test_native_runner.py:272`** — `test_run_native_agent_reports_the_adapter_resolved_prediction`

> ``trace.final_prediction`` is recorded further downstream, at the single point every inference path converges (``DefaultDataAdapter.run_inference``), not here [...]

## Why I believe this is stale rather than something I'm missing

1. `AgentTrace` (`evalscope/api/agent/trace.py:87`) declares only `framework`, `strategy`, `environment`, `max_steps`, `trial_id`, `total_usage`, `events`.
2. It declares no `model_config`, so under pydantic v2 the attribute cannot be assigned dynamically either.
3. `grep -rn "final_prediction" evalscope/ tests/` returns exactly those three comments — no definition, no assignment anywhere in the repo.

## Why it matters

`_emit_submit`'s docstring exists specifically to stop a reader conflating *what the loop observed* with *what actually gets scored* — a distinction that matters for adapters like SWE-bench that resolve their own artifact out of the trajectory. Following the pointer leads nowhere, so the warning loses its second half exactly where a new adapter author needs it.

As far as I can tell the resolution actually happens in `AgentAdapter._on_inference` via `_extract_final_answer`, landing in `InferenceResult.output.completion` (`evalscope/api/benchmark/adapters/agent_adapter.py:269-273`), and `DefaultDataAdapter.run_inference` is indeed the convergence point the test docstring describes.

## Question

Was `final_prediction` removed, or is it planned but not yet implemented?

- **Removed** → I've opened a comments-only PR repointing the three references at `_extract_final_answer`.
- **Planned** → the comments are right and the missing field is the real gap; happy to help there instead, and I'll close the PR.

One note either way: the test docstring records genuinely useful history (*"an earlier revision wrote it in this runner and consequently left it unset for every adapter that assembles its own `InferenceResult` (SWE-bench among them)"*). I kept that paragraph and only corrected the reference — it seems worth preserving.

ChenCJ-io:
Closes #1695 (or supersedes it, depending on the answer there).

## What

`AgentLoop._emit_submit` and `_emit_implicit_submit` both tell the reader that the reported prediction "lives in `AgentTrace.final_prediction`", and `test_run_native_agent_reports_the_adapter_resolved_prediction` repeats it.

`AgentTrace` has no such field:

- `evalscope/api/agent/trace.py:87` declares only `framework`, `strategy`, `environment`, `max_steps`, `trial_id`, `total_usage`, `events`
- it declares no `model_config`, so under pydantic v2 the attribute cannot be set dynamically either
- `grep -rn "final_prediction" evalscope/ tests/` returned only those three comments — no definition, no assignment

This repoints them at `AgentAdapter._extract_final_answer`, which resolves the prediction into `InferenceResult.output.completion` (`agent_adapter.py:269-273`).

## Why it's worth the churn

`_emit_submit`'s docstring exists precisely to stop a reader conflating *what the loop observed* with *what actually gets scored* — the distinction that matters for adapters like SWE-bench resolving their own artifact from the trajectory. Following the pointer currently leads nowhere, so the warning drops exactly where a new adapter author needs it.

## What I deliberately did not touch

The test docstring's history note — *"an earlier revision wrote it in this runner and consequently left it unset for every adapter that assembles its own `InferenceResult` (SWE-bench among them)"* — is real and useful, so I kept the paragraph and changed only the stale name. `DefaultDataAdapter.run_inference` is still named there; it is genuinely the convergence point.

## Scope

Comments only. No behaviour change, no signature change, no test logic change.

## Checks

- `ruff check evalscope/api/agent/loop.py tests/agent/test_native_runner.py` → All checks passed
- `ruff format --diff` → already formatted (`tests/**` is excluded from the formatter by `pyproject.toml`)

## If I read this wrong

If `final_prediction` is planned rather than removed, then the docstrings are correct and the missing field is the actual gap — say so on #1695 and I'll close this and help with the field instead.

### DIFF
--- a/evalscope/api/agent/loop.py
+++ b/evalscope/api/agent/loop.py
@@ -359,9 +359,10 @@ class AgentLoop:
         """Record the strategy's own termination signal.
 
         ``final_answer`` here is what the *loop observed*, not the prediction
-        that gets reported: that is resolved after the loop returns and lives
-        in ``AgentTrace.final_prediction``. The two coincide for the sentinel
-        protocol but must not be conflated.
+        that gets reported: that is resolved after the loop returns by
+        ``AgentAdapter._extract_final_answer`` and lands in
+        ``InferenceResult.output.completion``. The two coincide for the
+        sentinel protocol but must not be conflated.
         """
         self._dbg(
             ctx,
@@ -381,7 +382,8 @@ class AgentLoop:
         The loop deliberately does not publish a ``final_answer`` here: it has
         no answer of its own, and a malformed turn's ``raw_text`` is incidental
         prose rather than a submission. Only a short preview is kept, for
-        diagnosis; the reported prediction is ``AgentTrace.final_prediction``.
+        diagnosis; the reported prediction is resolved after the loop returns
+        by ``AgentAdapter._extract_final_answer``.
         """
         malformed = parsed.outcome is TurnOutcome.MALFORMED
         self.trace.add_event(
--- a/evalscope/api/agent/loop.py
+++ b/evalscope/api/agent/loop.py
@@ -359,8 +359,8 @@ class AgentLoop:
         """Record the strategy's own termination signal.
 
         ``final_answer`` here is what the *loop observed*, not the prediction
-        that gets reported: that is resolved after the loop returns by
-        ``AgentAdapter._extract_final_answer`` and lands in
+        that gets reported: that is resolved after the loop returns by the
+        adapter's ``_extract_final_answer`` hook and lands in
         ``InferenceResult.output.completion``. The two coincide for the
         sentinel protocol but must not be conflated.
         """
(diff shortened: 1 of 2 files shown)


Yunnglin (approved):
LGTM — corrected the prediction-resolution docstrings; local checks and CI pass.

PR #1698, opened 2026-09-04
fix(bridge): record upstream generate failures on the agent trace

ChenCJ-io:
## Problem

When `Model.generate_async` fails, the bridge writes an error frame to the agent client and a one-line WARNING to the log. It records nothing on `AgentTrace`.

The log is not the artifact anyone analyses afterwards — the trace is. So every attempt the agent client retried disappears from it.

### Evidence

**`BridgeTraceRecorder` has no failure-recording interface.** Its full set of record methods (`trace_recorder.py`):

```
record_anthropic_turn   :72      record_run_start  :207
record_openai_turn      :87      record_run_end    :219
record_responses_turn   :102
record_gemini_turn      :118
```

**All eight `recorder.` call sites sit on the success path.** In `server.py` they are at `501 / 537 / 610 / 643 / 740 / 772 / 816 / 853`, every one a `record_*_turn` placed after `await generate_task` inside the `try`. A failure never reaches them.

**`_log_upstream_failure` only logs** (`server.py:965-971`):

```python
if cls_name in _UPSTREAM_BUSINESS_ERRORS:
    logger.warning(f'{tag} upstream {mode} {cls_name}: {str(exc)[:200]}')
    return
logger.exception(f'{tag} {mode} generate failed')
```

**The external path emits no `EventType.ERROR` at all.** Before this change:

```
$ grep -rn "EventType.ERROR" evalscope/agent/ evalscope/api/agent/
evalscope/api/agent/loop.py:345      # native: parse error
evalscope/api/agent/loop.py:433      # native: max_steps_exceeded
evalscope/api/agent/loop.py:448      # native: context overflow
```

Nothing under `evalscope/agent/external/`. Same `AgentTrace` schema, two paths, and only one of them records its failures.

**The one failure signal that does exist is at the wrong granularity.** `record_run_end(error=run_error)` (`adapter.py:194-198`) captures a CLI subprocess crash or timeout. When the CLI exits cleanly after its own retries absorbed five failed model calls, `run_error is None`.

### Why this matters

A sample that made 12 model calls, four of them rate-limited and retried, leaves 8 `MODEL_GENERATE` events on the trace.

- Retries are invisible, so latency / token / cost totals under-count systematically.
- The failure is a hole in the trace, indistinguishable from the agent choosing not to call the model.
- Comparative evaluation is the one that worries me: a model served by a flaky endpoint produces a trace that looks *cleaner* and *faster* than the same model on a stable one, because the failed attempts and their latency were never recorded. An infrastructure difference reads as a model-quality difference.

### Why I think this is a gap rather than a decision

`tests/agent/external/test_responses_error_paths.py` specifies the client-facing contract carefully — status code, event name, `payload['code']`, `sequence_number`, presence of `[DONE]`. It asserts nothing about the trace.

I did read the position in `_log_upstream_failure`'s docstring: upstream `APIError` subclasses are business as usual, the bridge is a faithful conduit, tracebacks for them are noise. I agree with that — but it is an argument about log verbosity, not about what belongs in the artifact. Downgrading a traceback to a WARNING and omitting the event from the structured trace are two different decisions, and only the first one is argued there. The costs are asymmetric too: log noise is an annoyance, while silently under-counted latency and tokens contaminate the eval conclusions themselves.

If the omission *is* deliberate, say so and I'll close this.

## Changes

**`BridgeTraceRecorder.record_turn_failure()`** appends an `ERROR` event at the step the attempt *would* have occupied, without advancing `_step`. The turn produced no assistant message, and the retry that replaces it claims that same step — so a failed attempt sits beside its retry instead of shifting every later step by one and breaking alignment with the native loop's numbering.

**`_handle_upstream_failure()`** wraps the existing `_log_upstream_failure`, leaving its severity logic untouched. All eight failure paths (json + stream × anthropic / openai-chat / responses / gemini) route through it.

**Three `test_bridge_lifecycle` fixtures move off `recorder=None`.** `TrialSession` types `recorder` as a `BridgeTraceRecorder`, not `Optional`, so `None` was never a production state and the handlers are entitled to use it on failure paths too. No assertion in those tests changed — I'd rather make the fixture faithful than add a defensive `if recorder is not None`, which would reintroduce exactly the silent-skip this PR removes.

`EventType.ERROR` already exists and the wire contract is untouched, so this is additive observability only.

## Testing

```
tests/agent/external/test_responses_error_paths.py
  on main:   3 failed, 2 passed
             "expected one ERROR event, got []"
             retry case: got [MODEL_GENERATE]   <- the failed attempt is simply absent
  with this: 5 passed

tests/agent/   316 passed (313 on main), same 7 pre-existing failures
               (test_sandbox_service, test_strategy_parsers — unrelated to the bridge)

ruff check + ruff format --diff   clean
```

- **Payload shape** `{source: 'upstream', mode, error, message}`. I did not add a `TraceSources` constant: that module describes itself as private to the AgentLoop, and `trace_recorder.py`'s existing payloads use literals.
- **Step semantics**: the failure shares a step with the retry that replaced it. Giving it its own step is the alternative, at the cost of shifting everything after it.
- **No i18n key.** `LoopErrorRow` looks up `trace.loopMessage.${msg}` and falls back to the raw string, so the UI will show `RuntimeError: ...`. Say the word if you'd rather have a translated code.
- **The Responses streaming `except` also catches a write failure after a successful generate**, where `source='upstream'` is not accurate. I kept the existing framing rather than splitting the block — the comments already inside it (`sequence_number: 1`, `code: 'api_error'`) assume the failure happened before any frame went out, and splitting would change wire behaviour.

## Out of scope

`asyncio.CancelledError` still bypasses these handlers. `test_stream_disconnect_waits_for_generation_task` asserts that propagation deliberately, so recording on cancellation would change tested behaviour. Left alone.

### DIFF
--- a/evalscope/agent/external/bridge/server.py
+++ b/evalscope/agent/external/bridge/server.py
@@ -486,7 +486,7 @@ class ModelProxyServer:
                 config=gen_config,
             )
         except Exception as exc:  # pragma: no cover - upstream-dependent
-            _log_upstream_failure(session, exc, mode='json')
+            _handle_upstream_failure(session, exc, mode='json', latency_ms=(time.monotonic() - started) * 1000)
             return web.json_response(
                 {
                     'error': {
@@ -538,7 +538,7 @@ class ModelProxyServer:
             _log_turn(session, output, latency_ms, mode='stream')
         except Exception as exc:  # pragma: no cover - upstream-dependent
             failure_handled = True
-            _log_upstream_failure(session, exc, mode='stream')
+            _handle_upstream_failure(session, exc, mode='stream', latency_ms=(time.monotonic() - started) * 1000)
             error_payload = {
                 'error': {
                     'type': 'api_error',
@@ -595,7 +595,7 @@ class ModelProxyServer:
                 config=gen_config,
             )
         except Exception as exc:  # pragma: no cover - upstream-dependent
-            _log_upstream_failure(session, exc, mode='json')
+            _handle_upstream_failure(session, exc, mode='json', latency_ms=(time.monotonic() - started) * 1000)
             return web.json_response(
                 {
                     'error': {
@@ -646,7 +646,7 @@ class ModelProxyServer:
             async for chunk in stream_responses_payload(payload):
                 await response.write(chunk)
         except Exception as exc:  # pragma: no cover - upstream-dependent
-            _log_upstream_failure(session, exc, mode='stream')
+            _handle_upstream_failure(session, exc, mode='stream', latency_ms=(time.monotonic() - started) * 1000)
             # Responses error frame shape per OpenAI SDK ``ResponseErrorEvent``:
             # event name ``error`` (NOT ``response.failed`` — that one requires
             # a fully-constructed ``Response`` object with id/created_at/output/usage,
@@ -724,7 +724,7 @@ class ModelProxyServer:
                 config=gen_config,
             )
         except Exception as exc:
-            _log_upstream_failure(session, exc, mode='json')
+            _handle_upstream_failure(session, exc, mode='json', latency_ms=(time.monotonic() - started) * 1000)
             return web.json_response(
                 {
                     'error': {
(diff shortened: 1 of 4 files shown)


Yunnglin (approved):
LGTM. Client disconnects are now correctly separated from upstream generation failures, regression coverage is in place, and all CI checks pass.

PR #1701, opened 2026-09-04
fix(mcp): preserve initialization errors and support service stdio

### LINKED ISSUE #1693: MCP over HTTP transport always fails with CancelledError; stdio unusable in service mode
## Summary

`agent_config.mcp_servers` with `type: "http"` never works: every sample fails with
`concurrent.futures._base.CancelledError` before any tool call is made. The *same*
config with `type: "stdio"` works fine.

Separately, `stdio` transport does not work under `evalscope service`, so in service
mode **neither** MCP transport is usable.

## Reproduction

Only needs `mock_llm` — no real model or third-party API required.

```bash
PY=python  # your evalscope env

# A) stdio  -> succeeds
$PY -m evalscope.cli.cli eval \
  --model mock --eval-type mock_llm \
  --datasets deepsearchqa --limit 1 \
  --judge '{"strategy":"rule"}' \
  --agent-config '{"mode":"native","tools":["submit"],"mcp_servers":[
      {"type":"stdio","name":"fetch","command":"'"$PY"'","args":["-m","mcp_server_fetch"]}]}' \
  --work-dir ./repro_stdio

# B) http  -> every sample fails with CancelledError
#    (point the url at any StreamableHTTP MCP server)
$PY -m evalscope.cli.cli eval \
  --model mock --eval-type mock_llm \
  --datasets deepsearchqa --limit 1 \
  --judge '{"strategy":"rule"}' \
  --agent-config '{"mode":"native","tools":["submit"],"mcp_servers":[
      {"type":"http","name":"x","url":"http://127.0.0.1:9199/mcp","timeout":120}]}' \
  --work-dir ./repro_http
```

A ends with `Finished evaluation for mock on ['deepsearchqa']`.
B ends with `score computed from 0/1 successful samples; 1 errored`.

## Traceback (case B)

```
File "evalscope/api/benchmark/adapters/agent_adapter.py", line 248, in _on_inference
    result: AgentLoopResult = run_agent_loop(
File "evalscope/api/agent/runner.py", line 121, in run_agent_loop
    return AsyncioLoopRunner.run(_run())
File "evalscope/utils/asyncio_runtime.py", line 340, in run
    return handle.run_sync(coro, timeout=timeout)
File "evalscope/utils/asyncio_runtime.py", line 222, in run_sync
    return future.result(timeout=timeout)
concurrent.futures._base.CancelledError
```

Note it is `CancelledError`, not `TimeoutError` — the future is already cancelled when
`result()` is called, i.e. the coroutine never got to run on the owned loop. My guess is
that the MCP SDK's `AsyncExitStack` / anyio task-group lifecycle conflicts with the
generation lifecycle of `AsyncioLoopRunner`, but I have not isolated it further.

Failure is immediate (all log lines share the same second) and the MCP server receives
no requests at all, so the connection is never established.

`stateless=True` vs `stateless=False` on the server side makes no difference. Concurrency
makes no difference either (`--eval-batch-size 1` and `48` both fail).

## Second, related problem: stdio does not work under `evalscope service`

Submitting the stdio variant through `POST /api/v1/eval/invoke` fails with:

```
File "asyncio/unix_events.py", line 818, in _start
    self._proc = subprocess.Popen(
File "subprocess.py", line 1745, in _get_handles
    errwrite = stderr.fileno()
io.UnsupportedOperation: fileno
```

The evaluation runs in a `multiprocessing.spawn` child of the Flask service, where
`sys.stderr` has been replaced by a non-file logging object, so `stdio_client` cannot
spawn the MCP server process.

Combined with the HTTP issue above, this means **MCP tools cannot be used at all in
service mode** — only `CLI + stdio` works.

## Environment

- evalscope 1.11.1 (also reproduced the HTTP issue on the CLI path, so not service-specific)
- mcp 1.29.1
- Python 3.11.15, Linux

## Impact / why this may have gone unnoticed

No built-in benchmark uses `agent_config.mcp_servers`: `mcp_atlas` talks to its own
gateway with plain synchronous `requests` (`/enabled-servers`, `/list-tools`,
`/call-tool`), and other agentic benchmarks provide tools via `build_tools()`. So this
only affects users attaching their own external MCP tools — which is exactly what the
DeepSearchQA adapter docstring recommends:

> To evaluate a web-capable agent, set `TaskConfig.agent_config` and attach the
> search/fetch tools that should be available to the model.

Happy to test a patch.

git-jxj:
MCP HTTP initialization failures currently surface as `concurrent.futures.CancelledError`, hiding the underlying connection or HTTP error. Handle cancellation during initialization and unwind the async exit stack in the owning task, preserving the original exception and releasing transport resources.

Service workers also capture stderr with `StringIO`, which prevents lazily imported MCP stdio clients from starting subprocesses (`UnsupportedOperation: fileno`). Capture stderr in a temporary text file with a real file descriptor while retaining diagnostics in task errors.

Fixes #1693.

Validation:
- Regression tests fail before the fix and pass afterward, covering connection failures, HTTP 401/500, external cancellation, repeated stdio MCP sessions in spawned workers, and stderr forwarding.
- 60 related agent/service tests passed with MCP 1.29.1; CI smoke test and `make lint` passed.
- Actual `POST /api/v1/eval/invoke` checks with `mock_llm`, a local single-sample dataset, and MCP 1.29.0: stdio and healthy HTTP complete successfully; failed HTTP connections retain `httpx.ConnectError` in the error response.

Healthy HTTP transport worked before the fix as well; this change corrects failure reporting and cleanup rather than claiming every HTTP endpoint was broken.

### DIFF
--- a/evalscope/api/agent/mcp/client.py
+++ b/evalscope/api/agent/mcp/client.py
@@ -9,7 +9,8 @@ from __future__ import annotations
 
 import datetime
 from contextlib import AsyncExitStack
-from typing import Any, Dict, List, Optional
+from types import TracebackType
+from typing import Any, Dict, List, Optional, Type
 
 from evalscope.utils.import_utils import check_import
 from evalscope.utils.logger import get_logger
@@ -100,21 +101,26 @@ class MCPServer:
             await self._session.initialize()
             logger.info(f'MCPServer[{self.name}]: initialised')
             return self
-        except Exception:
-            # Make sure we don't leak any process / socket on partial init failure.
-            await self._stack.aclose()
-            self._stack = None
-            self._session = None
+        except BaseException as exc:
+            # MCP transport failures cancel the initializing task. Exit their
+            # task groups in this same task to surface the underlying error.
+            await self.__aexit__(type(exc), exc, exc.__traceback__)
             raise
 
-    async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
+    async def __aexit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> bool:
         if self._stack is not None:
             try:
-                await self._stack.aclose()
+                return await self._stack.__aexit__(exc_type, exc_val, exc_tb)
             finally:
                 self._stack = None
                 self._session = None
                 logger.debug(f'MCPServer[{self.name}]: closed')
+        return False
 
     async def list_tools(self) -> List[Any]:
         """Return the raw ``mcp.types.Tool`` list from the server."""
--- a/evalscope/service/utils/process.py
+++ b/evalscope/service/utils/process.py
@@ -1,10 +1,11 @@
 import contextlib
-import io
 import multiprocessing
 import queue
 import sys
+import tempfile
 import threading
 import traceback
+from typing import Iterator, TextIO
 
 from evalscope.config import TaskConfig
 from evalscope.perf.arguments import Arguments as PerfArguments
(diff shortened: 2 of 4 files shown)


Yunnglin (approved):
LGTM。已核对 MCP 初始化异常保留、资源清理及 service stdio stderr 捕获逻辑；本地相关测试与现有 CI 均通过，建议 squash merge。

Issue #1578, opened 2026-08-14
AIME judge verdict set by the word "Yes" anywhere in judge text; AIR-Bench scores taken from the last two bare numbers


AUTHENSOR:
## Summary

Two benchmark adapters parse LLM judge output with positional or substring
heuristics that are not bound to the judge's actual decision. In the AIME
adapters, the correctness verdict is True whenever the word "Yes" appears
anywhere in the judge's free-text response, so a wrong answer is scored correct
when the judge says "Yes" while explaining why the answer is wrong. In the
AIR-Bench Chat adapter, the two scores are taken from the last two bare numbers
in the judge output, so a candidate can choose its own scores by ending its
answer with score-shaped numbers that the judge then quotes or echoes. Both
results feed the benchmarks' reported metrics (`acc` for AIME, `judge_score`
and `win_rate` for AIR-Bench Chat).

## Affected code

Commit: 31ccc8c02671832df1f0a4aba64689c850898c95

1. `evalscope/benchmarks/aime/aime_adapter.py:183`

   ```python
   is_correct = bool(re.search(r'\bYes\b', judge_response, re.IGNORECASE))
   ```

   This line sits in `llm_match_score` (lines 171-194). The boolean sets
   `score.value['acc']`, the reported accuracy metric. `AIME25Adapter`
   (line 241) and `AIME26Adapter` (line 289) subclass `AIME24Adapter` without
   overriding the method, so aime24, aime25, and aime26 are all affected. The
   judge prompt (lines 18-77) asks for a bare "Yes" or "No" with no rationale,
   but the parser does not enforce that contract; it searches the entire
   response.

2. `evalscope/benchmarks/air_bench/air_bench_chat_adapter.py:384-402`
   (`_extract_judge_scores`)

   The parser returns the first line containing exactly two bare numbers in
   the 0-10 range, and otherwise falls back to `slash_nums[-2:]` or
   `nums[-2:]` over the whole judge output. `_judge_pair` (lines 356-382)
   consumes `nums[0]` and `nums[1]` as the two assistants' scores with only a
   1..10 range check. `llm_match_score` (lines 284-354) averages these into
   `judge_score`, the benchmark's primary metric, and `win_rate`. The judge
   prompt template (lines 42-54) embeds the candidate's answer verbatim, which
   gives the candidate a channel into the parsed text.

## PoC

Two standalone scripts (stdlib only, exit 0) download the genuine adapter
[...]

PR #1702, opened 2026-09-04
fix(eval): correct judge rendering, metric scoring, and bootstrap RNG

### LINKED ISSUE #1697: Follow-up to #1578: judge prompt placeholder injection ({gold} splice); unseeded bootstrap CI bounds; metric crash recorded as zero with SUCCESS status; allow_inclusion grades by substring
## Summary

Four defects at pin 3c52bd0 (extends #1578, which was fixed in #1588).

### 1. Judge prompt placeholder injection

Sequential str.replace over the whole prompt means a model answer containing the literal `{gold}` splices the reference answer into the predicted-answer section. A wrong answer scores 1.0.

### 2. Unseeded bootstrap confidence bounds

The arena bootstrap resamples with no seed; three fresh-process runs on identical battles produce different 97.5 percent bounds (0.6875, 0.7203, 0.7828).

### 3. Metric crash recorded as zero with SUCCESS status

The adapter's except-handler writes value=0 and keeps the default SUCCESS status, so a crashing metric plugin counts as a wrong answer in the mean denominator, contradicting the project's own ScoreStatus contract.

### 4. Accuracy with allow_inclusion grades by Python `in`

A list reference gives full credit for one-of-several required letters while the complete answer scores zero; alias lists become exact-match case-sensitive; string references become substring.

All four reproduce with deterministic offline stubs. Available as PRs with tests.

git-jxj:
Fixes #1697.

This fixes four scoring correctness issues: model text can expand judge-template placeholders, arena confidence intervals depend on unrelated global RNG activity, metric exceptions count as successful zero scores, and inclusion accuracy uses different matching rules for strings and lists.

- Render judge placeholders in one pass over the original template. Inserted questions, predictions, and references remain literal, including `{gold}`, JSON, and backslashes.
- Give `general_arena` bootstrap its own RNG, seeded from `TaskConfig.seed`. Fixed seeds reproduce confidence intervals regardless of global RNG consumption; explicit `None` retains nondeterministic sampling without modifying global RNG state.
- Omit failed metric keys and mark fully failed samples `EXCLUDED` or partially failed samples `DEGRADED`. Preserve valid metrics and their denominators through judge fallback and LLM recall, including a missing primary metric and equivalent metric names. When the judge recovers a completely failed rule, align equivalent metric names with the rule declaration so `acc` and `accuracy` do not split the cross-sample aggregation denominator.
- Match one complete, normalized acceptable answer for `allow_inclusion`. Lists remain alternatives (including MMLU-Redux single-choice answers); string references become a single alternative, so `York` no longer matches `New York`. Case and surrounding whitespace follow the existing normalization rules, and blank predictions do not receive credit.

The bootstrap issue is global-RNG coupling, not an absence of seeding in the standard entry point: `run_task` already defaults to seed 42. This does not change `arena_hard`, whose current aggregation does not call bootstrap. The AIME/AIR-Bench parsing defects in #1578 were already fixed by #1588 and subsequently migrated to JSON contracts in #1601; those adapters are unchanged here.

### Compatibility

`general_arena`, `mmlu_redux`, and `trivia_qa` move to evaluation version `v1.1`. A shared scoring version is also included in native cache fingerprints so old scores cannot bypass the common judge and metric fixes. This invalidates existing native score-cache identities, including benchmarks outside these three: use `rerun_review=True` to reuse predictions and recompute reviews. Legacy fingerprints remain calculated with their original algorithm for provenance tracking.

### Validation

Four of the 10 new cross-sample aggregation cases fail against `e0400d03`: equivalent judge and rule names produce duplicate accuracy aggregates with one sample each instead of one aggregate with two samples.

The results below are from the current revision unless marked as earlier. Regression tests were run against the old behavior and rerun after each fix. All judge responses use offline doubles.

- Python 3.13: **281 passed** across judge contracts/execution, adapter scoring, evaluation versioning, metrics/aggregators, and arena bootstrap:
  ```sh
  python -m pytest tests/api/judge tests/api/test_metric_failure_status.py tests/api/test_default_data_adapter.py tests/api/test_scoring_policy.py tests/api/test_evaluation_versioning.py tests/metrics tests/benchmark/test_arena_bootstrap.py -q --tb=short
  ```
- Earlier Python 3.12 focused regressions: **64 passed, 3 skipped**; the three real-ELO adapter cases were skipped because scikit-learn is unavailable in that environment, and passed under Python 3.13.
- `python -m pytest tests/cli/test_all.py::TestRun::test_ci_lite -v -s -p no:warnings`: **1 passed**, using the mock model.
- `make lint`: **passed**.

### DIFF
--- a/evalscope/api/benchmark/adapters/default_data_adapter.py
+++ b/evalscope/api/benchmark/adapters/default_data_adapter.py
@@ -12,7 +12,7 @@ from evalscope.api.messages import ChatMessage, ChatMessageSystem, ChatMessageUs
 from evalscope.api.metric import AggScore, SampleScore, Score
 from evalscope.api.model import Model, ModelOutput
 from evalscope.api.registry import get_aggregation, get_metric
-from evalscope.constants import HubType, JudgeStrategy
+from evalscope.constants import HubType, JudgeStrategy, ScoreStatus
 from evalscope.report import Report, ReportGenerator
 from evalscope.utils import get_logger
 
@@ -651,16 +651,19 @@ class DefaultDataAdapter(DataAdapter):
         )
 
         # Calculate scores for each configured metric
+        metric_failed = False
         for metric in self.metric_list:
             try:
                 if isinstance(metric, str):
                     metric_name = metric
-                    metric_scorer = get_metric(metric)  # Get metric implementation from registry
-                    metric_func = metric_scorer()  # Instantiate the metric scorer
+                    metric_args = {}
                 elif isinstance(metric, dict):
                     metric_name = list(metric.keys())[0]
-                    metric_cls = get_metric(metric_name)
-                    metric_func = metric_cls(**metric[metric_name])  # Initialize with parameters
+                    metric_args = metric[metric_name]
+                if score.main_score_name is None:
+                    score.main_score_name = metric_name
+                metric_cls = get_metric(metric_name)
+                metric_func = metric_cls(**metric_args)
                 metric_score = metric_func(
                     prediction=filtered_prediction,
                     reference=reference,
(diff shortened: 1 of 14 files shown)


Yunnglin (approved):
LGTM. Renaming the parametrized argument avoids the pytest-benchmark fixture collision; the full targeted regression suite (89 passed), pre-commit, and CI smoke test all pass locally.

PR #1703, opened 2026-09-04
feat(web): group same-model reports in the report list, display-only

### LINKED ISSUE #1599: Report list: group same-model reports for display, without merging report directories

Dhru1001:
Resolves #1599.

## Context

Our eval workflow runs one model against datasets incrementally — often as separate invocations (different dataset, different day) rather than one big run. The report list ends up with N unrelated rows per model, with no way to see a model's full picture without opening each report individually.

We tried physically merging report directories in #1570; @Yunnglin correctly closed it because merging mutates/deletes source evaluation runs, which conflicts with reports being immutable audit records. This PR is the display-only alternative discussed on #1599.

## What this does

- `GET /api/v1/reports?group_by=model` rolls same-model report-list items into one row each, entirely in memory, over metadata `_build_report_meta` already loads for the flat list. No report is read, written, moved, or merged.
- Each group carries `children`: every constituent report exactly as the flat list describes it, with its own `run_id`, `model_id`, and `primary_metrics`. A group never shows a fabricated combined score.
- If a dataset appears in more than one of the model's reports, there is no silent "winner" — both stay visible under their respective reports when the group is expanded.
- The response is a new `ListReportsGroupedResponse` (`reports: ReportGroup[]`), kept fully separate from the existing `ListReportsResponse` — the flat endpoint's contract is untouched.
- `ReportGroup` reuses the flat row's field names (`timestamp`, `dataset_name`, `num_samples`) where the meaning matches, so the existing sort machinery and `ReportsTable`/`ReportCard` render the expanded children with no group-aware branching. `refs` (`run_id/model_id` per child) feeds the existing multi-report Compare flow directly.
- The report list UI adds a "Group by model" toggle that renders these as expandable rows.

## Test plan

- [x] Backend: `pytest tests/report/test_report_endpoints.py` (15 passed)
- [x] Frontend: `npx tsc -b` clean, `npx vitest run` (463 passed)
- [x] `npm run contracts:check` — generated API contracts match the Pydantic models, no drift


Yunnglin:
Thanks for the display-only grouping approach. One small issue: for same-model reports from different datasets, "Compare all" opens a comparison with no common datasets, so it has no useful result. Could we remove this button for now, or show it only when the reports share a dataset? The grouping itself looks good.

Dhru1001:
Good catch, thanks. Pushed a fix: `Compare all` now only shows when at least two of the group's reports actually share a dataset (same intersection `ComparePage`'s own table already computes), instead of just checking `report_count >= 2`. For a group whose reports come from disjoint datasets, the button no longer appears. Added a test covering both cases.

Dhru1001:
@Yunnglin

Yunnglin (approved):
LGTM. Verified the grouped report UI and Compare-all behavior in a real browser; all checks are green.

PR #1712, opened 2026-09-08
feat(agent): validate tool-call arguments against the advertised schema

ChenCJ-io:
## Problem

`ToolExecutor` dispatches every call straight to its handler. Nothing checks `call.function.arguments` against the `ToolInfo.parameters` the model was shown, so a call with a missing required field or a wrong type is only noticed if the handler happens to crash on it.

That leaves two states indistinguishable in the results: *the model picked the right tool but wrote malformed arguments*, and *the tool itself failed*. For a run whose purpose is to measure tool use, the first one is the interesting signal and it is currently invisible.

The schemas are already in hand — `AgentContext.tools` carries them and the loop passes them to `model.generate`. The only place that validates against them today is the single-turn `FunctionCallAdapter` (`general_fc`, `k2_verifier`, …); the multi-turn loop never does.

```
$ grep -rn "jsonschema" evalscope/api/agent/ evalscope/agent/
(nothing)
```

## Change

`NativeAgentConfig.validate_tool_arguments`, **default `False`**. When enabled, `ToolExecutor` checks each call before dispatching it and refuses the ones that violate their schema.

- **`api/tool/utils.py`** — `validate_tool_arguments(call, tool)` validates with `jsonschema` against `tool.parameters.model_dump(exclude_none=True)`, i.e. exactly the schema the model was given, and returns the first violation as text with its property path (`"'abc' is not of type 'integer' (at 'limit')"`). A schema `jsonschema` cannot compile is treated as unconstrained, so a broken declaration never blocks a well-formed call.
- **`api/agent/tool_executor.py`** — `ToolExecutor(tool_infos=..., validate_arguments=...)`. A violating call is not executed; it returns a `ToolCallError(type='parsing')` through the same path an unregistered tool already takes. The model receives the violation as its observation and can correct itself on the next turn; the `TOOL_RESULT` trace event carries `error='parsing'`.
- The flag is threaded through `run_agent_loop`, `run_native_agent` and `AgentLoopAdapter._on_inference`. MCP tools are covered for free because `run_agent_loop` hands the merged `ToolInfo` list to the executor.
- One row added to the `NativeAgentConfig` table in `docs/{en,zh}/user_guides/agent/native.md`.

Nothing changes unless the flag is set, so existing benchmark scores are unaffected.

- **Executor rather than strategy.** The executor is already the place that refuses an unregistered tool without executing it, and putting the check there covers every strategy and MCP tools in one spot instead of once per strategy. The cost is threading one boolean through three call sites.
- **Reuses `ToolCallError(type='parsing')`** rather than adding an enum member, so no public contract changes and existing consumers of `TOOL_RESULT.payload.error` keep working.
- **Extra properties are rejected too.** `ToolParams` advertises `additionalProperties: false`, so an unexpected key is a violation of the contract the model was shown. Say the word if you would rather only enforce `required` + types.
- **Bridge mode is out of scope.** There the CLI executes its own tools and EvalScope only observes, so the same flag cannot refuse a call; annotating bridge `TOOL_CALL` events with a validation result would be a separate, observation-only change.

## Testing

```
tests/agent/test_tool_argument_validation.py   16 passed
  - validator: valid / missing required / wrong type / unexpected key / uncompilable schema
  - executor:  refuses without dispatching, passes valid calls through, leaves
               schemaless tools alone, off by default, unknown tool still 'unknown'
  - loop:      model gets the violation, corrects itself, handler only ever sees the
               corrected call, trace shows error=['parsing', None]
  - plumbing:  flag reaches the executor from run_agent_loop / run_native_agent /
               AgentLoopAdapter, and defaults to off without a NativeAgentConfig

tests/agent                                    336 passed, 21 skipped, same 7 pre-existing
tests/cli/test_all.py::TestRun::test_ci_lite   passed (mock_llm, 6m39s)
ruff 0.16.4 check + format --diff              clean
```

Sanity-checked against the built-in tools with the flag on: `bash {'command': 'ls'}` and `{'command': 'ls', 'timeout': 30}` pass, `{'timeout': '30'}` and `{}` are refused with the messages above.

## Note on CI

`main` is import-broken at `c31f0f9` (#1702 imports `evalscope.metrics.semantics.identity`, which #1705 moved to `naming.py`), so `unit-tests` fails here for a reason unrelated to this PR. #1711 fixes that one line; I ran the suites above with it applied locally.

### DIFF
--- a/docs/en/user_guides/agent/native.md
+++ b/docs/en/user_guides/agent/native.md
@@ -61,6 +61,7 @@ Most-used `NativeAgentConfig` fields:
 | `environment_extra` | Agent environment constructor options | Docker image, timeout, mounts, and other environment-specific settings |
 | `max_steps` | Max iterations per sample | Math/QA `5-10`, code fixes `100+` |
 | `skills_dir` | Optional host Agent Skills directory | EvalScope makes the skills available before the agent loop starts |
+| `validate_tool_arguments` | Reject tool calls whose arguments violate the JSON schema advertised to the model; the model sees the violation as a `parsing` tool error and can retry | `False` (default); enable when the run should measure whether the model produces schema-valid calls |
 | `kwargs` | Strategy kwargs | Most commonly `{'system_prompt': '...'}` to steer the model |
 
 Available `strategy` values:
--- a/docs/zh/user_guides/agent/native.md
+++ b/docs/zh/user_guides/agent/native.md
@@ -61,6 +61,7 @@ EvalScope 自动给模型注入 `python_exec` 工具定义，在 Docker 容器�
 | `environment_extra` | Agent 环境构造参数 | Docker 镜像、超时、挂载等环境专属设置 |
 | `max_steps` | 单样本最大轮数 | 数学/QA `5-10`，代码修复 `100+` |
 | `skills_dir` | 可选的本机 Agent Skills 目录 | EvalScope 会在 agent loop 开始前让 skills 可用 |
+| `validate_tool_arguments` | 拒绝参数不符合下发 JSON schema 的工具调用；模型会收到 `parsing` 类型的工具错误并可重试 | `False`（默认）；需要衡量模型能否产出合规参数时开启 |
 | `kwargs` | 策略参数 | 最常用 `{'system_prompt': '...'}` 引导模型行为 |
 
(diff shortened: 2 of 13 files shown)


Yunnglin (approved):
LGTM

PR #1715, opened 2026-09-08
feat(agent): stall the loop when the model repeats the identical tool call

ChenCJ-io:
## Problem

`AgentLoop` budgets two ways of failing to make progress:

| failure | budget |
|---|---|
| turn produces no tool call | `max_nudges` (2) |
| turn breaks the output protocol | `max_parse_error_nudges` (3) |

A model that keeps issuing one *valid* call is neither. Every turn is a well-formed `ACT` turn, and an `ACT` turn resets both counters, so the streak never accumulates and the only remaining exit is `max_steps` — with the same tool executed on every step along the way.

That budget is not small:

| benchmark | `max_steps_default` |
|---|---|
| `swe_bench_*_agentic`, `swe_bench_pro`, `gdpval`, `job_bench` | 250 |
| `mcp_atlas` | 100 |
| `gaia`, `wide_search`, `researchrubrics` | 50 |

A model that starts repeating at step 3 of SWE-bench spends the other 247 on real model calls and real sandbox executions. The trace it leaves is shaped exactly like a sample that genuinely needed every step, so "this task was hard" and "the model got stuck immediately" are indistinguishable without reading the turns by hand.

## Change

`NativeAgentConfig.max_repeated_tool_calls`, **default `None` (off)**. It counts consecutive turns whose tool calls match the previous turn's by name *and* arguments. On reaching the threshold the turn is converted into a malformed one: the tools are not executed, the model is told what it keeps repeating, and a model that never changes course ends on the existing malformed-nudge budget.

Measured on a stuck model with `max_steps=250`:

| threshold | model calls | tool executions | terminal reason |
|---|---|---|---|
| `None` | 250 | 250 | `max_steps_exceeded` |
| `3` | 6 | 2 | `parse_error_exhausted` |
| `5` | 8 | 4 | `parse_error_exhausted` |

Nothing changes unless the threshold is set.

- **In the loop, not in a strategy.** `FunctionCallingStrategy.max_tool_calls_per_turn` is the closest precedent and is strategy-level, but it is a per-turn rule while this one spans turns. Putting the guard in the loop covers all four strategies at once — including `swe_bench_toolcall`, which drives the 250-step benchmarks and would otherwise need its own copy.
- **Reuses the malformed-turn budget** rather than adding a third counter. Both mean "recoverable turn that made no progress", and separate counters would let either failure consume the other's retries — the reason `nudge_count` and `parse_error_nudge_count` were split in the first place.
- **The streak lives on `AgentContext`,** next to those two counters and owned by the loop for the same reason given in their docstrings. It cannot be derived from `ctx.messages`: `swe_bench_backticks` synthesises its `bash` call while parsing a fenced block, so its assistant messages carry no `tool_calls` to compare.
- **Arguments are part of the signature.** A poll that advances an offset, or a retry that changes a parameter, reads as progress. Call *order within* a turn is not part of it, so `{A, B}` and `{B, A}` count as the same turn.
- **`TraceSources.REPEATED_TOOL_CALLS`** marks the `ERROR` event so the guard is distinguishable from a genuine parse error without matching message text. `_emit_parse_error` gained a `source` parameter defaulting to the existing `'parse'`; no existing payload changes.
- **Thresholds below 2 are rejected** by a validator — `1` would stall the first call of every kind.

## Testing

```
tests/agent/test_repeated_tool_calls.py   16 passed
  streak semantics — identical calls trip at the threshold; changed arguments,
                     changed tool name and an idle turn each reset it; order
                     within a turn is irrelevant
  loop behaviour   — the guarded turn does not reach the handler, the model is
                     told and recovers, a model that never changes ends on the
                     nudge budget rather than max_steps, off by default,
                     genuine parse errors keep source='parse'
  plumbing         — the threshold reaches the loop from run_agent_loop,
                     run_native_agent and AgentLoopAdapter, and defaults to None

tests/agent                              357 passed, 21 skipped, same 7 pre-existing
tests/cli/test_all.py::TestRun::test_ci_lite   passed
ruff 0.16.4 check + format --diff        clean
```

## Follow-up, not in this PR

The `ERROR` event this emits is a natural input for a trace-derived "degenerate loop rate" alongside tool-error rate and nudge counts, which today are recorded per sample but never aggregated into the report. Happy to open an issue for that separately if it sounds useful.

### DIFF
--- a/docs/en/user_guides/agent/native.md
+++ b/docs/en/user_guides/agent/native.md
@@ -62,6 +62,7 @@ Most-used `NativeAgentConfig` fields:
 | `max_steps` | Max iterations per sample | Math/QA `5-10`, code fixes `100+` |
 | `skills_dir` | Optional host Agent Skills directory | EvalScope makes the skills available before the agent loop starts |
 | `validate_tool_arguments` | Reject tool calls whose arguments violate the JSON schema advertised to the model; the model sees the violation as a `parsing` tool error and can retry | `False` (default); enable when the run should measure whether the model produces schema-valid calls |
+| `max_repeated_tool_calls` | Stall the loop when the model issues the identical call (same name and arguments) this many times in a row, instead of letting it burn every remaining step | `None` (default, off); `3`-`5` for long budgets such as SWE-bench |
 | `kwargs` | Strategy kwargs | Most commonly `{'system_prompt': '...'}` to steer the model |
 
 Available `strategy` values:
--- a/docs/zh/user_guides/agent/native.md
+++ b/docs/zh/user_guides/agent/native.md
@@ -62,6 +62,7 @@ EvalScope 自动给模型注入 `python_exec` 工具定义，在 Docker 容器�
 | `max_steps` | 单样本最大轮数 | 数学/QA `5-10`，代码修复 `100+` |
 | `skills_dir` | 可选的本机 Agent Skills 目录 | EvalScope 会在 agent loop 开始前让 skills 可用 |
 | `validate_tool_arguments` | 拒绝参数不符合下发 JSON schema 的工具调用；模型会收到 `parsing` 类型的工具错误并可重试 | `False`（默认）；需要衡量模型能否产出合规参数时开启 |
+| `max_repeated_tool_calls` | 模型连续发出完全相同的调用（工具名与参数都相同）达到该次数时打断本轮，避免把剩余步数全部烧掉 | `None`（默认关闭）；SWE-bench 这类大步数预算建议 `3`-`5` |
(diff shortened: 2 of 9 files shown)


Yunnglin:
Thanks for the well-tested contribution. We are closing this PR because repeated identical tool calls alone do not prove that an agent has made no progress: polling and retries may intentionally issue the same call while external state changes.

We do not yet have production traces or rate data showing this is a material failure mode, and the disabled default still performs signature-tracking work. We prefer to preserve the current loop semantics rather than impose a task-independent policy on normal agent behavior.

If concrete traces demonstrate a problematic pattern, a future proposal with tool- or benchmark-specific repeat semantics and measured impact would be welcome.

Issue #1685, opened 2026-08-31
swe_bench_verified_agentic 测试hang住


ltm920716:
## 自查清单

在提交 issue 之前，请确保您已完成以下步骤:
- [x] 我已仔细阅读了[相关使用说明文档](https://evalscope.readthedocs.io/zh-cn/latest/get_started/parameters.html)
- [x] 我已查看了[常见问题解答](https://evalscope.readthedocs.io/zh-cn/latest/get_started/faq.html)
- [x] 我已搜索并查看了现有的 issues，确认这不是一个重复的问题

## 问题描述

swe_bench_verified_agentic 测试hang住

## EvalScope 版本（必填）
v1.10.0

## 使用的工具
- [ ] Native / 原生框架
- [ ] Opencompass backend
- [ ] VLMEvalKit backend
- [ ] RAGEval backend
- [x] Perf / 模型推理压测工具
- [ ] Arena / 竞技场模式

## 执行的代码或指令

```
evalscope eval \
  --model $MODEL \
  --api-url $OPENAI_API_COMPAT_URL \
  --api-key $TOKEN \
  --eval-type openai_api \
  --datasets swe_bench_verified_agentic \
  --ignore-errors \
  --eval-batch-size 64 \
  --generation-config '{"max_tokens": 131072, "temperature": 1.0, "top_p": 0.95, "retries":3, "parallel_tool_calls":"True"}' \
  --enable-progress-tracker \
  --use-cache ./outputs
```

## 错误日志

```
[INFO:ms_enclave] [📦 dqj57qxe] 2.5.2
[INFO:ms_enclave] [📦 dqj57qxe] 5.2.0
[INFO:ms_enclave] [📦 dqj57qxe] 5.2.1
[INFO:ms_enclave] [📦 dqj57qxe] 5.2.2
[INFO:ms_enclave] [📦 dqj57qxe] 5.2.3
[INFO:ms_enclave] [📦 dqj57qxe] e856638ba Preparing release version 5.2.3 (#6190)
[INFO:ms_enclave] [📦 dqj57qxe] 5e8c47faa Preparing release version 5.2.3
[INFO:ms_enclave] [📦 dqj57qxe] dae238c9b Release version 5.2.2 (#6055)
[INFO:ms_enclave] [📦 dqj57qxe] b27ba9772 Preparing release version 5.2.2
[...]

PR #1717, opened 2026-09-08
fix(agent): stop waiting on a sandbox that never returns

ChenCJ-io:
## Problem

Fixes the hang reported in #1685.

`EnclaveAgentEnvironment.exec` hands the deadline to ms_enclave's `shell_executor` and then awaits it with nothing bounding the wait:

```python
result = await handle.execute_tool(
    'shell_executor',
    {'command': shell_argv, 'timeout': timeout_s},
)
```

If the sandbox layer never answers — an unkillable command, a wedged exec channel, a lost reply — that `await` never returns. The sample blocks its worker for the rest of the run, with no traceback and no progress.

`LocalAgentEnvironment` does not have this problem: it wraps the subprocess in `asyncio.wait_for` and returns `timed_out=True` on expiry. Only the enclave path trusts the layer below it to honour the timeout it was handed.

In #1685 the model ran `sphinx-quickstart` inside the sandbox — an interactive CLI whose log tail is `* Please enter some text.` repeated forever, i.e. a prompt loop that never terminates on its own.

### Reproduction

No Docker, no ms_enclave, just a stubbed handle:

```python
class HangingHandle:
    sandbox_id = 'stub'
    async def execute_tool(self, name, args):
        await asyncio.sleep(3600)      # the sandbox never answers

env = EnclaveAgentEnvironment.__new__(EnclaveAgentEnvironment)
env._handle = HangingHandle(); env._timeout = 60.0; env._interpreter = ['bash', '-c']
await asyncio.wait_for(env.exec(['/bin/bash', '-c', 'sphinx-quickstart'], timeout=1.0), timeout=5)
```

```
STILL RUNNING after 5.0s despite exec(timeout=1.0)
```

## Change

`exec` stops waiting at `timeout + _TIMEOUT_GRACE_S` (30s) and returns a timed-out `ExecResult` carrying an explanatory `stderr`. The bash tool renders it as `[TIMEOUT]`, the model sees a failed command, and the run moves on.

The grace is waited out on purpose rather than cutting at the command deadline: ms_enclave reports `TIMEOUT` **together with whatever the command printed before its own deadline**, and that output is the more useful answer. The backstop only decides how long to wait for it before giving up.

## Scope

This bounds the damage. It does not explain why a sandbox would fail to enforce the timeout it was given — that still needs the reporter's `instance_id` and `pip show ms-enclave`, which I asked for on the issue.

Two related gaps I found while reading this code are **not** in this PR, both noted on #1685:

- `exec` accepts `input=` and never uses it, so stdin is silently dropped in a sandbox. `runners/mock.py` passes it; `runners/codex.py` already carries a comment working around it.
- `LocalAgentEnvironment` runs with `stdin=None`, so a child inherits the parent's stdin. Under a pty an interactive command blocks on it for the whole tool timeout (verified locally). `DEVNULL` would make those fail fast.

Happy to send either separately.

## Testing

```
tests/agent/test_t2_environment.py   3 new cases, 57 passed, 10 skipped
  - a handle that never returns is abandoned near the grace window rather
    than hanging (fails on main)
  - a sandbox reporting its own TIMEOUT slightly late is still waited for,
    so its partial output survives
  - a legitimately long command gets timeout + grace, not the grace alone

tests/agent                          344 passed, 21 skipped, same 7 pre-existing
tests/cli/test_all.py::TestRun::test_ci_lite   passed
ruff 0.16.4 check + format --diff    clean
```

### DIFF
--- a/evalscope/agent/environments/enclave.py
+++ b/evalscope/agent/environments/enclave.py
@@ -19,6 +19,7 @@ shared ``atexit`` hook.
 
 from __future__ import annotations
 
+import asyncio
 import re
 import shlex
 import time
@@ -132,6 +133,13 @@ class EnclaveAgentEnvironment(AgentEnvironment):
     """
 
     name: str = 'enclave'
+
+    #: Wall-clock grace allowed beyond the requested command timeout before the
+    #: environment stops waiting on the sandbox layer itself. The sandbox owns
+    #: the deadline and returns a ``TIMEOUT`` status with whatever output the
+    #: command produced, which is the better answer; this margin only decides
+    #: how long to wait for that answer before giving up on it.
+    _TIMEOUT_GRACE_S: float = 30.0
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


Yunnglin:
Closing this PR rather than merging it.

The proposed `asyncio.wait_for` fallback stops the caller from waiting, but does not reliably cancel the operation already submitted to `SandboxService`; the sandbox can therefore continue running in the background and remain unsafe to reuse.

The original report cannot be reproduced with the supported `ms-enclave>=0.0.8` Docker timeout path. A backend-specific fix should be made only with a reproducible failing case and must own cancellation and cleanup at the sandbox layer.

PR #1719, opened 2026-09-08
fix(agent): forward exec stdin into the sandbox instead of dropping it

ChenCJ-io:
## Problem

`EnclaveAgentEnvironment.exec` accepts `input=` and never reads it — the only occurrence of the name in the file is the signature:

```python
async def exec(
    self,
    cmd: List[str],
    *,
    cwd: Optional[str] = None,
    input: Optional[str] = None,  # noqa: A002 - mirrors ABC signature
    ...
```

`input` is part of the `AgentEnvironment` contract and `LocalAgentEnvironment` honours it, so a caller that pipes a payload gets it in one environment and silently loses it in the other.

- `runners/mock.py` passes `input=task.instruction` and would run against empty stdin in a sandbox.
- `runners/codex.py` carries a comment routing around the gap: *"EnclaveAgentEnvironment (whose `exec` does not forward `input=` to the underlying shell executor)"*.

The base class documents the policy for the analogous case — a subclass that cannot honour `env` *"should raise `NotImplementedError` rather than silently dropping the request"*. Neither branch was taken for `input`.

Found while investigating #1685; unrelated to the hang itself.

## Change

ms_enclave's `shell_executor` takes a command and no stdin, so the payload is piped in by the shell that already runs the rendered command:

```
printf %s <quoted payload> | ( <command> )
```

The subshell is load-bearing. The renderer prepends env exports and a `cd`, so without it a rendered `cd /w && cat` would feed `cd` rather than `cat`:

```
printf %s payload | ( export FOO=bar; cd /w && cat )
```

Supplying stdin this way is shell syntax, so a non-shell `interpreter` now raises `NotImplementedError` rather than dropping the payload a second time.

The two `codex.py` comments describing the gap are dropped since it no longer holds. The positional prompt stays as it is — it keeps the invocation visible in run logs.

## Testing

Verified in a real bash across every shape the renderer produces — bare command, cwd prefix, env prefix, pipelines, and payloads containing newlines, quotes, backticks, `$` and `%`.

```
tests/agent/test_t2_environment.py   5 cases, 4 of them fail on main
  input is rendered as a pipe
  the pipe survives cwd and env prefixes
  the rendered command runs in a real bash and round-trips a payload with
    newlines and shell metacharacters
  no input leaves the command unchanged
  a non-shell interpreter raises instead of dropping the payload

tests/agent   346 passed, 21 skipped, same 7 pre-existing failures
              (test_sandbox_service, test_strategy_parsers: optional deps)
ruff 0.16.4 check + format --diff   clean
```

I don't have ms_enclave installed, so the sandbox side is covered by asserting the exact command string handed to `shell_executor` plus executing that same string in a real shell. If there is a reason `shell_executor` would reject a pipeline, say so and I'll switch to raising `NotImplementedError` instead.

### DIFF
--- a/evalscope/agent/environments/enclave.py
+++ b/evalscope/agent/environments/enclave.py
@@ -79,6 +79,7 @@ def _render_command(
     interpreter: Sequence[str],
     cwd: Optional[str],
     env: Optional[Dict[str, str]],
+    stdin: Optional[str] = None,
 ) -> str:
     unwrapped_command = _unwrap_bash_c(cmd) if _interpreter_is_bash(interpreter) else None
     if unwrapped_command is not None:
@@ -92,6 +93,17 @@ def _render_command(
     if env:
         prefix = _render_env_exports(env)
         command = f'{prefix} {command}' if prefix else command
+    if stdin is not None:
+        if not _interpreter_is_bash(interpreter):
+            raise NotImplementedError(
+                f'EnclaveAgentEnvironment cannot supply stdin through interpreter {list(interpreter)!r}; '
+                'it is rendered as a shell pipeline. Use a bash interpreter or pass the payload in the command.'
+            )
+        # ms_enclave's shell_executor takes a command and no stdin, so the
+        # payload is piped in by the shell instead. The subshell keeps the
+        # pipe attached to the whole command: without it a rendered
+        # ``cd /w && foo`` would feed ``cd`` rather than ``foo``.
+        command = f'printf %s {shlex.quote(stdin)} | ( {command} )'
     return command
 
 
@@ -232,7 +244,7 @@ class EnclaveAgentEnvironment(AgentEnvironment):
         env: Optional[Dict[str, str]] = None,
     ) -> ExecResult:
         handle = await self._ensure_sandbox()
-        command = _render_command(cmd, interpreter=self._interpreter, cwd=cwd, env=env)
+        command = _render_command(cmd, interpreter=self._interpreter, cwd=cwd, env=env, stdin=input)
 
         # ms_enclave's shell_executor splits a bare string with no shell
         # wrapping; use an explicit interpreter so cd/&&/env-prefix/quoting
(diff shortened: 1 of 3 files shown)


Yunnglin:
Thanks for fixing the stdin path. One compatibility issue remains: `_render_command()` rejects every non-bash interpreter when `input` is set, but `['sh', '-c']` is already an accepted interpreter configuration and supports `printf %s payload | ( command )` as POSIX shell syntax.

Could we support at least `sh` and `bash` here, and add a real stdin round-trip test for `sh -c`? The current `python3 -c` rejection test is still useful for genuinely non-shell interpreters.

ChenCJ-io:
Good catch — you're right, and the gate was wrong for the reason you give: the rendering is POSIX syntax, not a bash extension, so keying it on `_interpreter_is_bash` refused interpreters that handle it perfectly well.

Pushed `ad3a2a7`:

- `_interpreter_is_posix_shell()` now gates the pipe, matching the interpreter basename against `ash / bash / dash / ksh / sh / zsh`. `ash` is in there for the busybox `sh` that slim images ship.
- `_interpreter_is_bash` keeps its own meaning — it gates `bash -c` unwrapping, which is a separate decision — and now shares a basename helper. `bash`, `/bin/bash` and `mybash` classify exactly as before.
- The `NotImplementedError` now names the accepted interpreters instead of just saying "bash".

On the tests:

- `test_rendered_input_pipe_runs_in_a_real_shell` is parametrized over `bash` and `sh`, so the round-trip really executes under `sh`. On CI `/bin/sh` is dash, which makes that the strict-POSIX case rather than another bash run.
- A new parametrized rendering case covers `sh`, `/bin/sh`, `dash` and `zsh`. Those assert the exact command handed to `shell_executor`, so they don't require the shells to be installed on the runner.
- The rejection case is parametrized over `python3 -c` and `node -e`, so genuinely non-shell interpreters stay loud.

I also checked the pipeline by hand under bash, sh, dash, ksh and zsh with a payload containing newlines, quotes, backticks, `$` and `%` — all five round-trip it intact.

Yunnglin (approved):
LGTM

PR #1722, opened 2026-09-09
fix(audio): preserve empty ASR transcripts when scoring WER

Excelius-Wang:
An ASR response containing `{"text": ""}` is currently treated as missing text. In Seed-TTS-Eval, the resulting exception is caught and recorded as WER=0, although an empty transcript against a nonempty reference should have WER=1.

Preserve empty primary transcripts and consult fallback fields only when the primary value is None. Add adapter-level coverage, field-precedence tests and error controls. Bump Seed-TTS-Eval's evaluation version to v1.1 because the scoring behavior changes.

Validation: the same expanded suite produces 4 failures and 24 passes with the baseline metric implementation, and 28 passes with the fix. The regressions cover adapter scoring, metric scoring, and empty primary text being overwritten by alternate fields. Missing/null fields, non-object JSON responses, and HTTP errors remain errors at the metric boundary. Applicable pre-commit hooks and `git diff --check` pass.

The scope is successful `/audio/transcriptions` responses with string text fields. HTTP responses are mocked; no production incident or live ASR response is claimed. The adapter's general error-to-zero fallback and empty Responses-protocol outputs are separate unresolved issues.

### DIFF
--- a/evalscope/benchmarks/seed_tts_eval/seed_tts_eval_adapter.py
+++ b/evalscope/benchmarks/seed_tts_eval/seed_tts_eval_adapter.py
@@ -30,6 +30,7 @@ PROMPT_TEMPLATE = (
 @register_benchmark(
     BenchmarkMeta(
         name='seed_tts_eval',
+        evaluation_version='v1.1',
         pretty_name='Seed-TTS-Eval',
         dataset_id='evalscope/Seed-TTS-Eval',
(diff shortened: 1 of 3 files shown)


Yunnglin (approved):
LGTM. The fix correctly preserves a valid empty ASR transcript and falls back only for None. The evaluation-version bump and regression coverage for field precedence, malformed payloads, and HTTP errors are appropriate; targeted tests and CI are green.

PR #1724, opened 2026-09-09
perf(web): cap dashboard's aggregated results table to 100 rows up front

Dhru1001:
## Summary
- `/dashboard`'s aggregated results table mounted every model x benchmark x metric row as table DOM on first paint with no cap. With enough models and benchmarks in the output directory, this makes the dashboard slow to render.
- Renders the first 100 rows up front, with a "show N more" control to reveal the rest in batches of 100. No data is hidden or paginated away server-side — everything already loaded is still there, this only staggers how much becomes DOM at once.
- Small result sets (the common case today) render exactly as before, with no visible change.

## Test plan
- [x] `npx tsc -b` clean
- [x] `npx eslint` clean
- [x] `npx vitest run` — 464/464 passing, including 2 new tests covering the cap and the "show more" reveal

### DIFF
--- a/evalscope/web/src/components/dashboard/AggregatedResults.tsx
+++ b/evalscope/web/src/components/dashboard/AggregatedResults.tsx
@@ -51,6 +51,17 @@ const DESCENDING_FIRST: Record<SortKey, boolean> = {
   lastRun: true,
 }
 
+/**
+ * Rows rendered up front, and added per click of "show more".
+ *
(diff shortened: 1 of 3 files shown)


Dhru1001:
@Yunnglin

Yunnglin (approved):
LGTM. Verified the dashboard with 120 aggregated rows: the initial render is capped at 100 and exposes ‘Show 20 more (of 120)’. The focused component test suite also passes (18/18).

PR #1725, opened 2026-09-09
fix(service): stop 500ing on predictions whose agent trace carries token usage

Dhru1001:
## Summary
- The `.../predictions` endpoint's `AgentTrace` response contract (`evalscope/service/api_models/reports.py`) had no `total_usage` field, but the base `ApiResponseModel` is `extra='forbid'`.
- Any stored prediction whose agent trace recorded token usage — which `claw_eval`, `acebench`, and the generic agent loop (`evalscope/api/agent/loop.py`) all populate via `trace.total_usage` — fails validation and the endpoint 500s instead of returning the predictions.
- Adds a `TraceUsage` contract mirroring `ModelUsage`'s shape (`input_tokens`, `output_tokens`, `total_tokens`, cache/reasoning token fields) and wires it onto `AgentTrace.total_usage`, then regenerates the frontend TS contracts.

## Test plan
- [x] `pytest tests/service/test_api_contracts.py` — 15/15 passing, including an extended fixture that reproduces the exact shape from the crash (`total_usage` alongside `input_tokens`/`output_tokens`)
- [x] `ruff check` / `ruff format --check` clean
- [x] `npm run contracts:check` — generated contracts match, no drift
- [x] `npx tsc -b` clean

### DIFF
--- a/evalscope/service/api_models/reports.py
+++ b/evalscope/service/api_models/reports.py
@@ -234,9 +234,11 @@ class TraceUsage(ApiResponseModel):
 
 
 class AgentTrace(ApiResponseModel):
+    framework: Optional[str] = None
     strategy: Optional[str] = None
     environment: Optional[str] = None
     max_steps: int
+    trial_id: Optional[str] = None
     events: List[AgentTraceEvent]
     total_usage: Optional[TraceUsage] = None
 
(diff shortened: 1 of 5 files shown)


Yunnglin (approved):
LGTM. The response contract now explicitly covers the persisted AgentTrace fields (`framework`, `trial_id`, and `total_usage`) while preserving strict validation. The generated frontend contracts and route-level regression test verify that `/predictions` accepts the real runtime serialization shape without silently relaxing the API boundary. CI is green.

PR #1727, opened 2026-09-11
fix(evaluator): preserve multi-alias targets for inclusion-based scoring

Dhru1001:
## Bug

`TaskState.target` always returns `Target.text`, which joins every accepted answer alternative into a single string with **no separator**:

```python
@property
def text(self) -> str:
    return ''.join(self.target)
```

For a benchmark like `trivia_qa`, whose samples set `target` to a list of answer aliases (`record['answer']['aliases'] + record['answer']['normalized_aliases']`), this collapses e.g. `["Oprah Winfrey", "Oprah", "Winfrey, Oprah Gail", ...]` into one unmatchable blob: `"Oprah WinfreyOprahWinfrey, Oprah Gail..."`.

`Accuracy(allow_inclusion=True)` (used by `trivia_qa` and `mmlu_redux`) is explicitly written to accept either a single string or a list of alternatives:

```python
alternatives = [reference] if isinstance(reference, str) else reference
```

but every call site passes `task_state.target`, which is always a `str` - so `alternatives` is always `[the_joined_blob]`, and a correct prediction can never exact-match it. Every sample in `trivia_qa` scores 0 regardless of whether the model's answer was right, because there's nothing left to match against.

Confirmed on a real run: the model answered "Of which African country is Niamey the capital?" with `ANSWER: Niger` (correct), and the sample still scored 0 - the gold reference used for grading was the fully-concatenated alias blob shown above.

## Fix

Return the full list of alternatives from `TaskState.target` when a sample has more than one, instead of always flattening to `Target.text`. Single-target samples (the common case across the rest of the benchmark suite) are unaffected and still return a plain string, since `len(self._target) > 1` is false for them - so this is scoped to exactly the multi-alias case that was broken.

Checked every other call site that reads `task_state.target` for a `str`-only assumption (e.g. `json.loads(task_state.target)` in `researchrubrics`) - none of them ever set a multi-item target, so they're untouched by this change.

### DIFF
--- a/evalscope/api/evaluator/cache.py
+++ b/evalscope/api/evaluator/cache.py
@@ -1,7 +1,7 @@
 import copy
 import os
 import uuid
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from pydantic import BaseModel, ConfigDict, Field, ValidationError, model_validator
 
@@ -431,8 +431,9 @@ class ReviewResult(BaseModel):
     index: int
     """Index of the sample that was reviewed."""
 
-    target: Optional[str] = None
-    """Expected/target answer for the sample, if available."""
+    target: Optional[Union[str, List[str]]] = None
+    """Expected/target answer for the sample, if available. A list when the
+    sample accepts more than one alias (see TaskState.target)."""
 
     messages: List[ChatMessage] = Field(default_factory=list)
     """Full chat message history exchanged during evaluation."""
--- a/evalscope/api/evaluator/state.py
+++ b/evalscope/api/evaluator/state.py
@@ -276,8 +276,20 @@ class TaskState:
         self._completed = completed
 
     @property
-    def target(self) -> str:
-        """The scoring target for this `Sample`."""
+    def target(self) -> Union[str, List[str]]:
+        """The scoring target for this `Sample`.
+
+        Returns the full list of accepted alternatives when the sample has more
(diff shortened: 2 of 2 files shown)


Dhru1001:
@Yunnglin

Yunnglin (changes requested):
Thanks for identifying the multi-alias scoring bug — the underlying issue is real, and preserving aliases for `Accuracy(allow_inclusion=True)` fixes the scoring path.

However, this currently breaks the predictions API for affected runs:

`TaskState.target` now returns `list[str]` → `ReviewResult.from_score_state()` persists that list → `_build_prediction_row()` assigns it to `Gold`, while `PredictionRow.gold` and the generated frontend contract still require `Gold: string`.

`get_predictions()` validates its successful response through `PredictionsResponse`, so a row with `Gold=['A', 'B']` raises a Pydantic validation error and the endpoint returns 500. I reproduced this directly with `PredictionsResponse.model_validate()`.

Could you please keep the serialized/display target as a string and expose aliases through a scoring-specific path, or update the complete API/frontend contract chain to support `string | string[]`? The former seems more scoped for this fix.

Please also add a regression test covering the full `TaskState -> scoring -> ReviewResult -> predictions endpoint` path for a multi-alias TriviaQA/MMLU-Redux sample.

Yunnglin (approved):
LGTM. The multi-reference scoring, cache migration, predictions API, and Dashboard rendering paths are covered by the updated tests and all required checks are green.

Dhru1001:
Thxxx @Yunnglin

PR #1731, opened 2026-09-13
fix(arena): validate review identities before pairing

elandesberg:
GeneralArena pairs saved reviews by row position. If candidate indices are `[0, 2]` and baseline indices are `[0, 1, 2]`, it can compare the candidate's answer to question 2 with the baseline's answer to question 1, then silently discard the remaining row. Native saved transcripts also contain generated responses, which must be excluded from the judge's question.

This change validates imported observations before constructing battles. Each shared dataset/subset must have matching, non-empty index sets, consistent supplied sample/repeat identities, and matching inputs. Errors identify the model, dataset/subset, and conflicting indices. Repeated generations retain their individual indices. Duplicate cache rows follow the existing resume convention: use the last saved row and log a warning.

For modern caches, input comparison preserves message roles and structured content while excluding per-run message IDs/metrics. Tool-call IDs are renamed consistently across declarations and references, preserving call/result relationships, function arguments, and provider payloads. Both comparison and the judge's question use only the input prefix before the first generated message; answers remain in their respective answer fields. Explicit legacy `input` rows retain rendered-text compatibility, and unmarked histories remain input because their generation boundary cannot be recovered. The arena evaluation version advances to `v1.2`, and both user guides document the import requirements and recovery steps.

Validation:

- `make lint` — passed.
- `pytest tests/benchmark/test_arena_review_pairing.py tests/benchmark/test_arena_bootstrap.py tests/evaluator/test_cache_manager.py tests/api/test_evaluation_versioning.py tests/api/judge -q` — 198 passed, including 48 native report/JSONL pairing cases.
- Four regression checks fail on the prior PR head: two expose generated content in the question, and two expose false mismatches from tool-call IDs. All pass after the fixes; negative controls still reject changed call/result links, arguments, and provider payloads.
- `pytest tests/cli/test_all.py::TestRun::test_ci_lite -v -s -p no:warnings` — passed with mock model inference.
- Two offline `run_task` evaluations on three local General-QA questions with `repeats=2` produced six correctly matched arena samples. Every question equals its original input without a generated answer. Removing one saved candidate observation caused the expected error before judging. Network connections were disabled for this reproduction.


elandesberg:
Addressed both review findings in 390a0573c47df09dffc599f99fd33d2ee3f206e0.

- Arena questions now use the same input prefix as identity comparison, excluding generated answers and tool trajectories while preserving input demonstrations.
- Tool-call IDs and their references are normalized consistently across runs, preserving call/result relationships. IDs inside arguments and opaque provider payloads remain unchanged.

The expanded pairing suite reproduced four failures before the fix and now passes all 48 cases. The broader regression suite passes 198 tests; `make lint` and the required CLI smoke test also pass. Two offline native evaluations produced six pairs, and all six judge questions matched the original input without generated answers.

Could a maintainer authorize the pending Lint Test and CI Tests Lite runs and review the updated commit? Both workflow runs currently report `action_required`.

Yunnglin:
Thanks for the thorough pairing validation. I think this still needs two fixes before merge:

1. `_review_input()` correctly strips messages from the first `source="generate"` entry onward for equality validation, but `process_review_item()` still builds `Question` from the full `review.messages_markdown`. For native review rows, that includes the generated assistant response, so the Arena judge can receive one model’s answer as part of the user prompt. Please construct the pair question from the same pre-generation input prefix, and add a regression assertion that generated content is absent from the resulting Arena sample input.

2. The structured comparison excludes the top-level message ID, but run-specific IDs can remain in tool-related fields such as `tool_call_id` and nested tool-call IDs. Two equivalent tool-call prompts from separate runs may therefore be rejected as mismatched. Please recursively normalize or exclude those runtime identifiers, with a regression case covering equivalent tool inputs that differ only in IDs.

With these two cases covered, the index-based pairing and fail-fast validation look good.

Yunnglin (approved):
LGTM. The follow-up correctly isolates judge questions from generated messages and normalizes tool-call runtime IDs without weakening input-structure validation. Targeted regressions and required CI checks pass.

PR #1744, opened 2026-09-14
feat(perf): add optional PD handoff metrics

gbdjxgp:
## Summary

  This PR adds optional metrics for analyzing PD-disaggregated inference serving in `evalscope perf`.

  In PD-disaggregated serving, the first generated token can be produced by the prefill node, while subsequent tokens are produced by the decode node.
  The interval between the first and second streamed output chunks can therefore reflect the P→D handoff path, but it also includes normal decode
  latency. To make this easier to inspect, this PR adds an opt-in metric group behind `--enable-pd-metrics`.

  New metrics:

  - `Steady ITL (ms)`
    - Excludes the first inter-output chunk interval from each request.
    - Intended to represent post-handoff steady decode inter-token/chunk latency.
    - Reported with avg / p50 / p99 / max and percentile JSON entries.

  - `PD Handoff Latency (ms)`
    - Uses the first inter-output chunk interval as a P-D handoff latency proxy.
    - Reported with avg / p50 / p99 / max and percentile JSON entries.

  - `PD Handoff Overhead (ms)`
    - Estimates extra handoff cost as:
      `max(0, first_inter_output_interval - median(subsequent_inter_output_intervals))`
    - This subtracts the steady decode baseline from the first post-TTFT interval.
    - Reported with avg / p50 / p99 / max and percentile JSON entries.

  The metrics are disabled by default, so standard non-PD perf benchmarks keep the existing TTFT / TPOT / ITL output unchanged.

  ## Validation

  Ran the requested perf command with `--enable-pd-metrics` against a local OpenAI-compatible mock service:

  ```bash
  .venv/bin/evalscope perf \
    --url "http://127.0.0.1:8000/v1/chat/completions" \
    --api openai \
    --model "mock-qwen" \
    --dataset random \
    --tokenizer-path "/data2/models/Qwen3-0.6B/" \
    --prefix-length 0 \
    --min-prompt-length 100 \
    --max-prompt-length 100 \
    --min-tokens 512 \
    --max-tokens 512 \
    --extra-args '{"ignore_eos": true}' \
    --parallel 10 \
    --number 20 \
    --connect-timeout 120 \
    --read-timeout 1200 \
    --log-every-n-query 100 \
    --enable-pd-metrics \
    --debug \
    --name "test" \
    --outputs-dir "./evalscope"

  Observed the new metrics in both the benchmarking summary and per-request metrics table:

  - Steady ITL (ms)
  - PD Handoff Latency (ms)
  - PD Handoff Overhead (ms)


Yunnglin:
Thanks for adding focused diagnostics for PD-disaggregated serving. Reusing the existing per-request inter-chunk timings and clearly treating the result as a proxy are both good design choices.

Before merge, could we please address the following?

1. \`--enable-pd-metrics\` currently gates only the final report output. The default-disabled path still derives \`pd_handoff_overhead\` with \`median(...)\` for every successful request and accumulates additional steady-ITL / PD values. Since this runs on the perf benchmarking hot path, please make the option gate the PD-specific derivation and accumulation as well, so standard benchmarks incur no extra CPU or memory cost and are not perturbed by an unused diagnostic feature.

2. Please add focused behavior tests for the new contract, beyond the semantics-registry coverage. In particular:
   - expected average and percentile values for a multi-interval stream;
   - a single-interval stream, where handoff latency is available but steady ITL / overhead are not;
   - non-stream or empty-interval requests;
   - disabled mode preserving the existing summary and percentile output contract.

Optional: \`summary_result()\` propagates the flag, but periodic logging and visualizer messages still call \`create_message()\` without it. Please either propagate the flag there too, or document that PD metrics are final-report-only.

gbdjxgp:
@Yunnglin 
Thanks for the review!

I have addressed the comments in the latest commit:

- Gated the PD-specific derivation and accumulation behind `--enable-pd-metrics`, so the default-disabled path no longer computes `pd_handoff_latency`, `pd_handoff_overhead`, or accumulates steady-ITL / PD-specific values.
- Added focused behavior tests for:
  - multi-interval stream average and percentile values;
  - single-interval stream;
  - non-stream / empty-interval requests;
  - disabled mode preserving the existing summary and percentile output contract.
- Also propagated `enable_pd_metrics` to the periodic logging and visualizer `create_message()` calls.

One point I’d like to confirm: do you think `Steady ITL` should remain as a separate metric, or would it be better to replace the existing `ITL` with steady ITL when `--enable-pd-metrics` is enabled?

Yunnglin (approved):
LGTM. The PD metric hot-path gate, behavioral coverage, and real vLLM end-to-end validation all look good.

PR #1745, opened 2026-09-14
fix(perf): forward tokenizer trust option for AgentX

Xiangyi1996:
AgentX replay cannot load tokenizers that require custom Python code because EvalScope does not expose AIPerf's `--tokenizer-trust-remote-code` option.

Add an explicit `tokenizer_trust_remote_code` scenario option, disabled by default, and forward it to AIPerf when enabled. Include an English usage example and tests covering the omitted, explicitly disabled, and enabled settings.

```bash
evalscope perf \
  --scenario '{"name":"agentx","mode":"smoke","tokenizer_trust_remote_code":true}' \
  --model YOUR_MODEL \
  --tokenizer-path YOUR_TOKENIZER_PATH_OR_ID \
  --url http://localhost:8000/v1/chat/completions \
  --parallel 1
```

Validation: `python -m pytest tests/perf/test_agentx.py -q` — 18 passed.

Repository checks: `make lint` and `pre-commit run --all-files` passed.

CLI smoke: `pytest tests/cli/test_all.py::TestRun::test_ci_lite -v -s -p no:warnings` — 1 passed.


Yunnglin (approved):
LGTM — AgentX tokenizer trust/revision and request-timeout forwarding are covered by targeted tests; unit-tests, perf-tests, and run-pre-commit all passed.

PR #1747, opened 2026-09-15
fix(perf): forward AgentX benchmark grace period

Xiangyi1996:
EvalScope's AgentX scenario does not expose the benchmark grace period, so users cannot configure how long to wait for outstanding responses after the sending duration ends. A longer per-request timeout does not extend this separate wait.

Add an optional `benchmark_grace_period` scenario field and forward it to AIPerf's existing `--benchmark-grace-period` option. The value sets the maximum wait after sending stops; it does not require waiting for the entire period when all in-flight requests have completed.

- Omitting the field preserves the existing default (30 seconds with the pinned AIPerf 0.12.0 dependency).
- Finite, nonnegative values are accepted, including zero.
- The sending duration and per-request timeout remain independently configured.
- English and Chinese documentation describe the setting and provide an example.

For example, allow up to 600 seconds for outstanding responses after the 1800-second sending period:

```bash
evalscope perf \
  --scenario '{"name":"agentx","benchmark_grace_period":600}' \
  --duration 1800 \
  --model YOUR_MODEL \
  --tokenizer-path YOUR_TOKENIZER_PATH_OR_ID \
  --url http://localhost:8000/v1/chat/completions \
  --parallel 1
```

Validation:

- AgentX unit tests: 30 passed, covering smoke and benchmark forwarding, zero, omitted values, invalid values, and independent request timeout/duration.
- `make lint`: passed.
- `pytest tests/cli/test_all.py::TestRun::test_ci_lite -v -s -p no:warnings`: passed using the mock model.


Xiangyi1996:
@Yunnglin  Could you take a look when you have a chance? This adds an optional AgentX grace-period setting while preserving the existing default. CI checks are passing. Thanks!

Yunnglin (approved):
LGTM. The change is focused, backward-compatible, and covered by tests.

PR #1751, opened 2026-09-19
fix(metrics): parse Chinese multi-select answers separated by 、 or /

feizhuzheng:
### Problem

`parse_answers_zh` drops all but the first label when a Chinese multi-select answer separates its labels with an ideographic comma (`、`) or a slash (`/`).

The English parser (`parse_answers`) already treats both as separators — this is covered by `test_parse_answers_multiple_correct_slash_and_ideographic_comma`. The Chinese path never gained the same handling: its plain-label pattern `_PLAIN_LABEL_ZH_RE` stops at `、`/`/`, and the multi-select branch only normalizes `，` to `,`. So a valid answer is truncated to a single label and scored as wrong.

### Reproduction

```python
from evalscope.utils.multi_choices import parse_answers, parse_answers_zh
# state has choices A/B/C/D, multiple_correct=True

parse_answers(state_en('ANSWER: A、C'),  multiple_correct=True)  # {'A', 'C'}  (English, correct)
parse_answers_zh(state_zh('答案：A、C'), multiple_correct=True)   # {'A'}       <- bug
parse_answers_zh(state_zh('答案：A/C'),  multiple_correct=True)   # {'A'}       <- bug
```

The bracketed form (`答案：（A，C）`) already works because it goes through `_BRACKETED_LABEL_RE`; only the plain/unbracketed form is affected.

### Fix

Bring the Chinese plain-label path to parity with the English one:

- `_PLAIN_LABEL_ZH_RE` now accepts `、` and `/` so the capture is not truncated at the first separator.
- The `multiple_correct` branch normalizes `、` and `/` to `,` before splitting (mirroring `parse_answers`).

Single-answer behavior is unchanged and stays consistent with the English parser (a stray multi-label answer like `答案：A、B` yields no answer in single-select mode, exactly as `ANSWER: A、B` does).

### Verification

```
$ pytest tests/test_multi_choices.py -q
25 passed
```

Added `test_parse_answers_zh_multiple_correct_slash_and_ideographic_comma`, mirroring the existing English separator test. Ran `ruff check` and `ruff format --check` on the changed files — both clean.


Yunnglin (approved):
LGTM — brings the Chinese multi-select parser (`parse_answers_zh`) to full parity with the English `parse_answers`, treating `、` and `/` as label separators. Reproduced the divergence on main and confirmed the fix; the added test mirrors the existing English separator test. unit-tests, run-pre-commit and perf-tests all pass.

PR #1752, opened 2026-09-20
fix(benchmarks): match archive members exactly in mvbench/videomme_v2 video lookup

feizhuzheng:
## Summary

Harden MVBench archive-member lookup with exact relative-path matching (with a basename fallback), aligned with tvbench, so a requested video resolves to the correct member instead of a substring-suffix collision.

## Scope

MVBench only. The Video-MME-v2 change has been dropped: its IDs are pinned to three fixed-length digits (`001`–`800`) and the `NNN.zip` archives only contain three-digit names, so `expected` can never be a proper suffix of another member — that half was purely defensive.

## Real collision (default `action_antonym` subset)

Contrary to my earlier note, the public MVBench mirror **does** reproduce a collision on the default run path. `DEFAULT_SUBSET_LIST = ['action_antonym']` pulls from `ssv2_video.zip`, whose members are Something-Something-v2 ids — non-zero-padded, variable-length numbers. The archive contains both `ssv2_video/9741.mp4` and `ssv2_video/209741.mp4`.

- Annotation requests `9741.mp4`.
- Old lookup: `name.endswith("9741.mp4")` matches both members, and `sorted(matches)[0]` returns `ssv2_video/209741.mp4` (lexically `'2' < '9'`).
- The wrong video is silently fed to the model; the item is scored incorrectly with no error raised.

The patched lookup resolves `9741.mp4` to `ssv2_video/9741.mp4`. Independently verified in the ModelScope mirror by @Yunnglin.

## Verification

- `pytest tests/benchmark/test_archive_member_lookup.py` (4 passed)
- Regression test uses the real names: `9741.mp4` must resolve to `ssv2_video/9741.mp4` when `209741.mp4` is also present. It fails against the pre-fix lookup and passes after.


Yunnglin:
Thanks for the defensive hardening and the test coverage. However, the PR currently lacks evidence that this affects the official datasets.

We downloaded and exercised the real archive-mode paths:
- MVBench `star.zip` uses five-digit, zero-padded names (`00001.mp4`–`00200.mp4`), and its annotations use the same format.
- Video-MME-v2 `001.zip` contains only `001.mp4`–`020.mp4`; valid video IDs are limited to 1–800.

Therefore, the `2.mp4`/`12.mp4` and `001.mp4`/`1001.mp4` cases are synthetic scenarios, not reproduced in the current public mirrors. Matching should be hardened only when there is an affected real archive/annotation pair or a supported local-data compatibility requirement.

Please provide a real reproducible collision case and its impact. Without that evidence, we should close this PR rather than merge it as a bug fix.

feizhuzheng:
Thanks for actually downloading and exercising the real archive-mode paths — that's the right bar, and it made me go re-verify against the public data instead of the constructed cases. Here's what I found.

**Video-MME-v2: you're right, there's no real collision.** `normalize_video_id` pins every id to exactly three digits (`001`–`800`) and the official `NNN.zip` archives only contain three-digit names, so `expected` can never be a proper suffix of another member. The `001.mp4`/`1001.mp4` case cannot occur on any supported path (local overrides still go through the same normalization). That half of the PR is purely defensive — happy to drop it if you'd prefer.

**MVBench: there is a real, reproducible collision on the default subset**, and it comes from a source I hadn't called out. `star.zip` uses 5-char alphanumeric hashes (`AJTDO.mp4`), so it's collision-free as you saw. But the **default** subset `action_antonym` (`DEFAULT_SUBSET_LIST = ['action_antonym']`) pulls from `ssv2_video.zip`, whose videos are Something-Something-v2 ids — **non-zero-padded, variable-length numbers**. That's where `endswith` bites.

Concrete case from the official data:
- `json/action_antonym.json` contains an annotation with `"video": "9741.webm"`.
- `video/ssv2_video.zip` contains **both** `ssv2_video/9741.webm` **and** `ssv2_video/209741.webm` (both are legitimately sampled videos in this subset).
- Old lookup: `name.endswith("9741.webm")` matches both, and `sorted(matches)[0]` returns `ssv2_video/209741.webm` (lexically `'2' < '9'`).

So for the QA item whose video is `9741.webm` ("Which one of these descriptions correctly matches the actions in the video?" → *Stuffing something into something*), the model is silently fed `209741.webm` (a *letting something roll up a slanted surface* clip). Wrong video in, item scored wrong — no error raised. The patched lookup resolves `9741.webm` correctly.

Minimal repro against the public MVBench release (verified via `OpenGVLab/MVBench` `json/action_antonym.json` + `video/ssv2_video.zip`; the `PKU-Alignment` ModelScope mirror the adapter defaults to ships the same SSv2 subset):

```python
import zipfile
z = zipfile.ZipFile("ssv2_video.zip")
members = [n for n in z.namelist() if not n.endswith("/")]
video_name = "9741.webm"   # from json/action_antonym.json
old = sorted(n for n in members if n.endswith("/"+video_name) or n.endswith(video_name))
print(old)          # ['ssv2_video/209741.webm', 'ssv2_video/9741.webm']
print(old[0])       # ssv2_video/209741.webm  <-- wrong video fed to the model
```

It's one collision out of the 200 items in the default subset, so the blast radius is small, but it's official public data and the default run path, not a synthetic scenario — hence I think the MVBench half is worth keeping as a bug fix. I'm glad to reframe the PR around MVBench only, drop the Video-MME-v2 change, and add a regression test using these exact names (`9741.webm` vs `209741.webm`) so it documents the real case. Let me know which shape you'd like and I'll update.

Yunnglin:
Thanks for the follow-up and the concrete MVBench reproduction. I independently verified the default `action_antonym` collision in the ModelScope mirror: the annotation requests `9741.mp4`, while the old suffix lookup selects `ssv2_video/209741.mp4` ahead of `ssv2_video/9741.mp4`. The MVBench fix is therefore justified.

Before merging, please narrow the PR to that verified scope:

1. Remove the Video-MME-v2 lookup change and its synthetic tests, since the supported three-digit ID/archive layout does not admit the proposed collision.
2. Replace or supplement the MVBench synthetic test with the real regression case: `9741.mp4` must resolve to `ssv2_video/9741.mp4` when `209741.mp4` is also present.
3. Update the PR description: the public MVBench mirror does contain this collision in the default `action_antonym` subset, so the current statement that the public mirrors do not reproduce a collision is inaccurate.

Once those changes are in place, this should be ready to merge.

feizhuzheng:
Thanks @Yunnglin, and thanks for independently reproducing the `action_antonym` collision — that's a great sanity check. I've narrowed the PR to exactly the three points you outlined:

1. **Reverted the Video-MME-v2 lookup change and dropped its synthetic tests.** With three-digit fixed-length IDs (`001`–`800`) and the matching `NNN.zip` layout, `expected` can't be a proper suffix of another member, so there was nothing real to fix there.
2. **Kept the MVBench fix and replaced the synthetic tests with the real regression.** The test now asserts that `9741.mp4` resolves to `ssv2_video/9741.mp4` when `209741.mp4` is also present in the archive (it fails against the old `endswith` lookup, which returns `209741.mp4`, and passes after the fix).
3. **Updated the PR description** to state that the public MVBench mirror does contain this collision in the default `action_antonym` subset, correcting the earlier inaccurate note, and scoped everything to MVBench-only.

`pytest tests/benchmark/test_archive_member_lookup.py` passes (4 tests). Let me know if you'd like anything else tightened.

PR #1753, opened 2026-09-20
fix(agent): prevent zero answers from falling back to assistant text

git-jxj:
## Summary

In an agent evaluation, a benchmark sample may have `0` as its expected answer. When a provider returns that answer as a JSON number through the `submit` tool, EvalScope previously treated it as missing because numeric zero is falsy in Python.

For example, with assistant content `"I am confident"` and submit answer `0`, the final prediction could become `"I am confident"` instead of the expected `"0"`. This can incorrectly score otherwise valid zero-answer samples.

This change preserves an explicitly submitted `0` as `"0"`. Fallback to assistant content now happens only when the answer is missing, null, or empty. The fix applies to the shared ReAct and Function Calling strategies.

## Tests

- Added three regression tests for numeric-zero submit answers
- 116 relevant agent tests passed
- Ruff check and format check passed

One unrelated SWE-bench test was excluded because its optional dependency is unavailable; the same failure occurs on the unchanged main branch.


Yunnglin (approved):
LGTM — default-mode submit extraction now preserves an explicit numeric zero as the final prediction, while strict tool-argument validation continues to enforce the string-only submit contract. Targeted agent tests and all CI checks pass.

PR #1754, opened 2026-09-20
fix(perf): preserve MMMU image placeholder order

git-jxj:
## Summary
- Preserve the `<image N>` / `<image_N>` positions from MMMU prompts when building multimodal performance requests.
- Keep the existing source-order fallback for rows whose prompt has no image placeholders.

## Root cause
`mmmu_multi_image` collected images in `image_1` ... `image_7` order and appended every image after one text block. MMMU prompts use numbered placeholders and may reference images in a different order, so the request could associate the wrong image with each placeholder and leave the placeholder text in the prompt.

## Verification
- Added a regression test that fails on `68d7b7df0a1abdc6794e3372ad5adcb6d2361e61` before the fix.
- `/root/vscode/evalscope/.venv/bin/pytest -q tests/perf/test_mmmu_multi_image.py` (7 passed)
- Ruff check and format check passed for the changed Python files.

## AI assistance disclosure
Prepared with AI assistance; the change, test, and repository-policy checks were reviewed before submission.


Yunnglin (approved):
LGTM — the MMMU image-placeholder ordering fix preserves the intended text-image binding, retains the no-placeholder fallback, and is covered by regression tests. unit-tests, perf-tests, and run-pre-commit are passing.

PR #1755, opened 2026-09-21
fix(ifbench): count keywords and person names on word boundaries

feizhuzheng:
## Summary

Two IFBench keyword/name checkers match **substrings** instead of whole words, so they diverge from the official IFBench reference and inflate scores on the `ifbench` benchmark:

- `KeywordsMultipleChecker` (`count:keywords_multiple`) uses `value.lower().count(keyword.lower())`. A required keyword like `art` is counted inside `start`, `smart`, `apart`, so the expected `(1, 2, 3, 5, 7)` occurrence counts are wrong whenever a keyword happens to be a substring of another word.
- `PersonNameCountChecker` (`count:person_names`) uses `if name in value`. Listed names match inside longer words — `Mia` inside `Miami`, `Ava` inside `Available`, `Leo` inside `Leonardo`, `Nora` inside `Norah` — so unrelated text is counted as person names.

Both produce false-positive scores on the official IFBench dataset.

## Fix

Match whole words with a word-boundary regex (`\b...\b`) in both checkers, which is exactly what the upstream IFBench reference implementation does (its `count:keywords_multiple` even carries the comment *"Use regex instead of str.count to avoid counting substrings"*). `re` is already imported in the module.

Bump IFBench `evaluation_version` from `v1.1` to `v1.2` so cached scores cannot silently cross the scoring-semantics change, consistent with the mechanism introduced in #1676.

## Why this matters (real impact)

These checkers run on the standard `ifbench` benchmark. For example:

- keyword `art`, response `"...the art of a start and a smart display..."` → naive `str.count("art") == 4`, so the checker fails a response that actually contains the whole word `art` once.
- name requirement of 1, response `"I flew to Miami last week."` → `"Mia" in value` is `True`, so a response mentioning no person at all is scored as satisfying the instruction.

Both are pure logic bugs, reproducible without any model server, and both currently disagree with the official IFBench scoring.

## Tests

```
pytest -q tests/benchmark/test_ifbench_instructions.py   # 16 passed
ruff check evalscope/benchmarks/ifbench/instructions.py evalscope/benchmarks/ifbench/ifbench_adapter.py tests/benchmark/test_ifbench_instructions.py   # All checks passed
ruff format --check evalscope/benchmarks/ifbench/instructions.py evalscope/benchmarks/ifbench/ifbench_adapter.py   # already formatted
```

Added regression tests for both checkers (substring inputs must now be rejected, whole-word inputs still accepted). Confirmed the new tests fail on the pre-fix code.


Yunnglin (approved):
LGTM。该 PR 针对 count:keywords_multiple 和 count:person_names 的子串误匹配，采用与上游一致的单词边界匹配，并通过 evaluation_version v1.2 隔离评分语义变化前后的缓存。回归测试覆盖了子串与完整词场景，CI 已通过。其余 IFBench 上游 verifier 对齐将另行跟进。

PR #1757, opened 2026-09-21
fix: give logger calls a placeholder for their extra positional arguments

Linxiushen:
## What

Five `logger.*` calls pass a value as an extra positional argument without a `%`-placeholder in the message:

| file | call |
|---|---|
| `evalscope/backend/rag_eval/clip_benchmark/utils/webdataset_convert.py` (×2) | `logger.info('Final dataset size:', nsamples)` |
| `evalscope/benchmarks/docmath/utils.py` | `logger.error('Error: ', prediction, type(prediction))` |
| `evalscope/benchmarks/image_edit/gedit/utils.py` | `logger.info('Now fixing: ', json_str)` |
| `evalscope/benchmarks/image_edit/gedit/utils.py` | `logger.info('Error: Cannot fix', json_str)` |

`evalscope.utils.logger.get_logger()` returns a standard `logging.Logger`, so the record is formatted with `msg % args`; with no placeholder that raises `TypeError: not all arguments converted during string formatting`. `logging` catches it, prints `--- Logging error ---` plus a traceback to stderr, and the message itself is never emitted:

```python
>>> log.info('Final dataset size:', 5)
--- Logging error ---
Traceback (most recent call last):
  ...
TypeError: not all arguments converted during string formatting
Call stack: ...
Message: 'Final dataset size:'
Arguments: (5,)
```

The docmath and gedit sites are on error-handling paths (`except` blocks), so the diagnostics meant to explain a failure were the ones being lost.

## Change

Add the missing `%s` placeholders. No behaviour change beyond the messages now actually being logged.

## Checks

- `ruff check` / `ruff format --check` on the three files: clean.
- No test added: the fix is five format strings; the failure mode above is standard `logging` behaviour.

---

_Written with Claude Code (AI-assisted) and submitted under the account owner's authorization._


Yunnglin (approved):
LGTM — completed the logger-format and metadata-file fixes; targeted tests, unit-tests, perf-tests, and pre-commit checks all pass.

PR #1758, opened 2026-09-21
fix(agent): remove the temporary HOME after codex/hermes/gemini-cli runs

Linxiushen:
## What

`CodexRunner`, `HermesRunner` and `GeminiCliRunner` call `_resolve_home()` on every `run()`, which returns `tempfile.mkdtemp(prefix='evalscope-<runner>-')` whenever no `home_override` is configured. The directory is handed to the CLI as `HOME` / `HERMES_HOME` and then forgotten — nothing removes it, so every evaluated sample leaves an `evalscope-codex-*` / `evalscope-hermes-*` / `evalscope-gemini-cli-*` directory behind (with whatever the CLI wrote into it: skills, config, caches).

The other three runners already handle this: `ClaudeCodeRunner`, `OpenCodeRunner` and `DeepSeekHarnessRunner` compute `owns_home_dir = home_dir is not None and self._home_override is None` and `shutil.rmtree(home_dir, ignore_errors=True)` in a `finally`. This PR applies the same pattern to the remaining three, so all six runners behave the same.

- `home_override=<path>` (user-supplied) and `home_override=''` (inherit the parent `HOME`) are untouched, exactly as in the existing runners.
- For codex the cleanup runs after the `--output-last-message` file has been read; for hermes after the `config.yaml` written into `HERMES_HOME` has been consumed by the CLI.
- The diff is mostly the re-indent under `try:`; reviewing with whitespace changes hidden shows the actual change (`import shutil`, the `owns_home_dir` line, `try:` / `finally:`).

## Tests

`tests/agent/external/test_runner_home_cleanup.py` (fake environment, no CLI needed), parametrized over the three runners:

- temporary HOME is removed after a successful run
- temporary HOME is removed when the CLI exits non-zero
- a user-supplied `home_override` directory is kept

On `main` the first two cases fail for all three runners (6 failed, 3 passed); with this change 9 passed. `ruff check` / `ruff format --check` clean on the changed files; `test_deepseek_harness_runner.py`, `test_opencode_runner.py`, `test_runner_skill_install.py`, `test_codex_runner.py`: 13 passed, 6 skipped (e2e opt-ins).

---

_Written with Claude Code (AI-assisted) and submitted under the account owner's authorization._


Yunnglin (approved):
LGTM. This PR makes the minimal change needed to clean up temporary HOME directories for the Codex, Hermes, and Gemini CLI runners while preserving inherited and user-supplied paths. The tests cover successful runs, CLI failures, and user-provided home directories.

